PR #3, opened 2022-12-24
Support for non-Float64 eltypes when using ForwardDiff

torfjelde:
See tests for example + things like

``` julia
x = ComponentVector(x = zeros(3))
∇ℓ = ADgradient(:ForwardDiff, ℓ; input=x)
```

also works.

### DIFF
--- a/src/AD_ForwardDiff.jl
+++ b/src/AD_ForwardDiff.jl
@@ -36,12 +36,12 @@ Wrap a log density that supports evaluation of `Value` to handle `ValueGradient`
 Keywords are passed on to `ForwardDiff.GradientConfig` to customize the setup. In
 particular, chunk size can be set with a `chunk` keyword argument (accepting an integer or a
 `ForwardDiff.Chunk`), and the underlying vector used by `ForwardDiff` can be set with the
-`input` keyword argument (accepting an `AbstractVector`).
+`x` keyword argument (accepting an `AbstractVector`).
 """
 function ADgradient(::Val{:ForwardDiff}, ℓ;
-                    input = zeros(dimension(ℓ)),
+                    x = zeros(dimension(ℓ)),
                     chunk = _default_chunk(ℓ),
-                    gradientconfig = _default_gradientconfig(ℓ, chunk, input))
+                    gradientconfig = _default_gradientconfig(ℓ, chunk, x))
     ForwardDiffLogDensity(ℓ, gradientconfig)
 end
 
--- a/src/AD_ForwardDiff.jl
+++ b/src/AD_ForwardDiff.jl
@@ -18,28 +18,30 @@ end
 
 _default_chunk(ℓ) = ForwardDiff.Chunk(dimension(ℓ))
 
-function _default_gradientconfig(ℓ, chunk::ForwardDiff.Chunk)
-    ForwardDiff.GradientConfig(Base.Fix1(logdensity, ℓ), zeros(dimension(ℓ)), chunk)
+function _default_gradientconfig(ℓ, chunk::ForwardDiff.Chunk, x::AbstractVector)
+    return ForwardDiff.GradientConfig(Base.Fix1(logdensity, ℓ), x, chunk)
 end
 
-function _default_gradientconfig(ℓ, chunk::Integer)
-    _default_gradientconfig(ℓ, ForwardDiff.Chunk(chunk))
+function _default_gradientconfig(ℓ, chunk::Integer, x::AbstractVector)
+    return _default_gradientconfig(ℓ, ForwardDiff.Chunk(chunk), x)
 end
 
 """
-    ADgradient(:ForwardDiff, ℓ; chunk, gradientconfig)
-    ADgradient(Val(:ForwardDiff), ℓ; chunk, gradientconfig)
+    ADgradient(:ForwardDiff, ℓ; input, chunk, gradientconfig)
+    ADgradient(Val(:ForwardDiff), ℓ; input, chunk, gradientconfig)
 
 Wrap a log density that supports evaluation of `Value` to handle `ValueGradient`, using
 `ForwardDiff`.
 
 Keywords are passed on to `ForwardDiff.GradientConfig` to customize the setup. In
 particular, chunk size can be set with a `chunk` keyword argument (accepting an integer or a
-`ForwardDiff.Chunk`).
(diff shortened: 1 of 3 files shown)


src/AD_ForwardDiff.jl
```diff
@@ -18,28 +18,30 @@ end
...
-function _default_gradientconfig(ℓ, chunk::Integer)
-    _default_gradientconfig(ℓ, ForwardDiff.Chunk(chunk))
+function _default_gradientconfig(ℓ, chunk::Integer, x::AbstractVector)
+    return _default_gradientconfig(ℓ, ForwardDiff.Chunk(chunk), x)
 end
 
 """
-    ADgradient(:ForwardDiff, ℓ; chunk, gradientconfig)
-    ADgradient(Val(:ForwardDiff), ℓ; chunk, gradientconfig)
+    ADgradient(:ForwardDiff, ℓ; input, chunk, gradientconfig)
```
devmotion, line 30: Should we unify the keyword arguments for ReverseDiff (`x`) and ForwardDiff?

torfjelde: Ah, didn't see that! Will do:)

src/AD_ForwardDiff.jl
```diff
@@ -18,28 +18,30 @@ end
...
 `ForwardDiff`.
 
 Keywords are passed on to `ForwardDiff.GradientConfig` to customize the setup. In
 particular, chunk size can be set with a `chunk` keyword argument (accepting an integer or a
-`ForwardDiff.Chunk`).
+`ForwardDiff.Chunk`), and the underlying vector used by `ForwardDiff` can be set with the
+`input` keyword argument (accepting an `AbstractVector`).
 """
 function ADgradient(::Val{:ForwardDiff}, ℓ;
+                    input = zeros(dimension(ℓ)),
```
devmotion, line 42: IMO it is nice to use default values of `nothing` since that clearly indicates if a user provided a custom value and makes it easy to dispatch later and define the default values in the downstream methods were they are actually needed (and possibly changed at some point).
So maybe similar to the ReverseDiff implementation use
```suggestion
                    input::Union{Nothing,AbstractVector} = nothing,
```

torfjelde: Hmm, though I see what you mean, it does add two additional methods (we need one for `chunk::Integer` and one for `chunk::FowardDiff.Chunk`).

torfjelde: And atm there's no need to know whether the user specified it or not, right?

torfjelde: Okay I think I found a "clean" way I'm happy with

src/AD_ForwardDiff.jl
```diff
@@ -16,30 +16,32 @@ function Base.show(io::IO, ℓ::ForwardDiffLogDensity)
...
+_default_gradientconfig(ℓ, chunk, ::Nothing) = _default_gradientconfig(ℓ, chunk, zeros(dimension(ℓ)))
+function _default_gradientconfig(ℓ, chunk, x::AbstractVector)
+    return ForwardDiff.GradientConfig(Base.Fix1(logdensity, ℓ), x, _chunk(chunk))
 end
 
 """
-    ADgradient(:ForwardDiff, ℓ; chunk, gradientconfig)
-    ADgradient(Val(:ForwardDiff), ℓ; chunk, gradientconfig)
+    ADgradient(:ForwardDiff, ℓ; input, chunk, gradientconfig)
+    ADgradient(Val(:ForwardDiff), ℓ; input, chunk, gradientconfig)
```
devmotion, line 31:
```suggestion
    ADgradient(:ForwardDiff, ℓ; x, chunk, gradientconfig)
    ADgradient(Val(:ForwardDiff), ℓ; x, chunk, gradientconfig)
```

src/AD_ForwardDiff.jl
```diff
@@ -16,30 +16,32 @@ function Base.show(io::IO, ℓ::ForwardDiffLogDensity)
...
 `ForwardDiff`.
 
 Keywords are passed on to `ForwardDiff.GradientConfig` to customize the setup. In
 particular, chunk size can be set with a `chunk` keyword argument (accepting an integer or a
-`ForwardDiff.Chunk`).
+`ForwardDiff.Chunk`), and the underlying vector used by `ForwardDiff` can be set with the
+`x` keyword argument (accepting an `AbstractVector`).
 """
 function ADgradient(::Val{:ForwardDiff}, ℓ;
+                    x = nothing,
```
devmotion, line 42: Maybe restrict types already here?
```suggestion
                    x::Union{Nothing,AbstractVector} = nothing,
```

src/AD_ForwardDiff.jl
```diff
@@ -16,30 +16,32 @@ function Base.show(io::IO, ℓ::ForwardDiffLogDensity)
...
                     chunk = _default_chunk(ℓ),
```
devmotion, line 43:
```suggestion
                    chunk::Union{Integer,ForwardDiff.Chunk} = _default_chunk(ℓ),
```

src/AD_ForwardDiff.jl
```diff
@@ -16,30 +16,32 @@ function Base.show(io::IO, ℓ::ForwardDiffLogDensity)
...
-                    gradientconfig = _default_gradientconfig(ℓ, chunk))
+                    gradientconfig = _default_gradientconfig(ℓ, chunk, x))
```
devmotion, line 44:
```suggestion
                    gradientconfig::ForwardDiff.GradientConfig = _default_gradientconfig(ℓ, chunk, x))
```
?

src/AD_ForwardDiff.jl
tpapp: I agree with the suggestion.

tpapp (changes requested):
Thanks for this PR, I am happy to merge once suggestions by @devmotion are implemented.

tpapp:
@devmotion: I am on holiday and will only check my emails occasionally in the next week or so, feel free to merge this and tag a release once your suggestions are implemented, so that TuringLang/AdvancedHMC.jl#301 is not held up.

torfjelde:
Great! Made the changes:)

devmotion (approved):
LGTM, thank you!

PR #7, opened 2023-02-06
Load only LogDensityProblemsAD and weak dependency in extension

devmotion:
See https://github.com/JuliaStats/LogExpFunctions.jl/pull/63.


tpapp (approved):
Thanks!

PR #8, opened 2023-02-13
Don't preallocate GradientConfig in ForwardDiff backend by default

### LINKED ISSUE #6: AD with ForwardMode is not thread-safe
When using multiple threads, the gradient computations disagree with the gradients one gets when using a single thread. Here's an MWE:

```julia
using ForwardDiff, LinearAlgebra, LogDensityProblems, LogDensityProblemsAD

struct LogDensityFunction{F}
    logp::F
    dim::Int
end
function LogDensityProblems.capabilities(::Type{<:LogDensityFunction})
    return LogDensityProblems.LogDensityOrder{0}()
end
LogDensityProblems.dimension(ℓ::LogDensityFunction) = ℓ.dim
LogDensityProblems.logdensity(ℓ::LogDensityFunction, x) = ℓ.logp(x)

dim = 10
n = 10^6
P = LogDensityFunction(x -> -sum(abs2, x) / 2, dim)
ℓ = ADgradient(:ForwardDiff, P)
x = randn(dim, n)

results_single = zeros(dim, n)
results_threaded = zeros(dim, n)

for j in 1:n
    results_single[:, j] = LogDensityProblems.logdensity_and_gradient(ℓ, x[:, j])[2]
end

Threads.@threads for j in 1:n
    results_threaded[:, j] = LogDensityProblems.logdensity_and_gradient(ℓ, x[:, j])[2]
end
```

On my machine (using 4 threads), I get the following result:
```julia
julia> norm(results_single .- results_threaded)
2588.220046954836
```

This appears to be the same issue as https://github.com/JuliaDiff/ForwardDiff.jl/issues/573, where the signature `ForwardDiff.gradient(f, x, cfg)`, which is used here, is shown to be not thread-safe. By comparison, ReverseDiff seems to be fine.

tpapp:
Fixes #6.

Changes:

- for the ForwardDiff backend, the default is now thread-safe. To preallocate a buffer, provide a type (this is a breaking API change, but I don't think anyone was using this feature anyway).
- suggest `copy` to make copies of AD'd densities. This is only necessary for ForwardDiff, but allows a consistent AD-agnostic interface for the caller in threaded code.
- documentation changes
- various incidental changes

### DIFF
--- a/ext/ForwardDiffExt.jl
+++ b/ext/ForwardDiffExt.jl
@@ -60,10 +60,12 @@ Keyword arguments:
 
 - `chunk` can be used to set the chunk size, an integer or a `ForwardDiff.Chunk`
 
-- `gradient_config_type` can be `nothing` (the default) or a type (eg `Float64`).
+- `gradientconfig` can be
+   1. `nothing` (the default), for creating a gradient config for each call,
+   2. a type (eg `Float64`), for preallocating a gradient config using that type, or
+   3. a `ForwardDiff.GradientConfig` value, which is used directly.
 
-   The latter preallocates and reuses a `ForwardDiff.GradientConfig` for that type. Note
-   that **this option is not thread-safe**. You can [`copy`](@ref) the results for
+   Note **the last two options are not thread-safe**. You can [`copy`](@ref) the results for
    concurrent evaluation:
    ```julia
    ∇ℓ1 = ADgradient(:ForwardDiff, ℓ; gradient_config_type = Float64)
@@ -73,9 +75,11 @@ Keyword arguments:
 function ADgradient(::Val{:ForwardDiff}, ℓ;
                     x::Union{Nothing,AbstractVector} = nothing,
                     chunk::Union{Integer,ForwardDiff.Chunk} = _default_chunk(ℓ),
-                    gradient_config_type::Union{Nothing,Type{T}} = nothing) where {T<:Real}
-    gradient_config = if gradient_config_type ≡ nothing
+                    gradientconfig::Union{Nothing,Type{T},ForwardDiff.GradientConfig} = nothing) where {T<:Real}
+    gradient_config = if gradientconfig ≡ nothing
         nothing
+    elseif gradientconfig isa ForwardDiff.GradientConfig
+        gradientconfig
     else
         isconcretetype(T) ||
             throw(ArgumentError("gradient_config_type needs to be a concrete subtype of Real."))
--- a/ext/ForwardDiffExt.jl
+++ b/ext/ForwardDiffExt.jl
@@ -77,10 +77,10 @@ function ADgradient(::Val{:ForwardDiff}, ℓ;
     gradient_config = if gradient_config_type ≡ nothing
         nothing
     else
-        T = gradient_config_type
-        (isconcretetype(T) && (T <: Real)) ||
+        S = gradient_config_type
+        (isconcretetype(T) && (S <: Real)) ||
             throw(ArgumentError("gradient_config_type needs to be a concrete subtype of Real."))
-        _make_gradient_config(T, ℓ, chunk)
+        _make_gradient_config(S, ℓ, chunk)
     end
     ForwardDiffLogDensity(ℓ, chunk, gradient_config)
 end
--- a/ext/ForwardDiffExt.jl
+++ b/ext/ForwardDiffExt.jl
@@ -91,7 +91,7 @@ function logdensity_and_gradient(fℓ::ForwardDiffLogDensity, x::AbstractVector)
     if gradient_config ≡ nothing
         gradient_config = _make_gradient_config(eltype(x), ℓ, chunk)
     end
-    result = ForwardDiff.gradient!(buffer, Base.Fix1(logdensity, ℓ), x, gradientconfig)
+    result = ForwardDiff.gradient!(buffer, Base.Fix1(logdensity, ℓ), x, gradient_config)
     _diffresults_extract(result)
 end
 
--- a/ext/ForwardDiffExt.jl
+++ b/ext/ForwardDiffExt.jl
@@ -77,10 +77,9 @@ function ADgradient(::Val{:ForwardDiff}, ℓ;
     gradient_config = if gradient_config_type ≡ nothing
         nothing
     else
-        S = gradient_config_type
-        (isconcretetype(T) && (S <: Real)) ||
+        isconcretetype(T) ||
             throw(ArgumentError("gradient_config_type needs to be a concrete subtype of Real."))
-        _make_gradient_config(S, ℓ, chunk)
+        _make_gradient_config(T, ℓ, chunk)
     end
     ForwardDiffLogDensity(ℓ, chunk, gradient_config)
 end
--- a/ext/ForwardDiffExt.jl
+++ b/ext/ForwardDiffExt.jl
@@ -18,12 +18,14 @@ end
 # Load DiffResults helpers
 include("DiffResults_helpers.jl")
 
-struct ForwardDiffLogDensity{L, C <: ForwardDiff.Chunk,
+struct ForwardDiffLogDensity{L, C <: ForwardDiff.Chunk, T <: Union{Nothing,ForwardDiff.Tag},
                              G <: Union{Nothing,ForwardDiff.GradientConfig}} <: ADGradientWrapper
     "supports zero-order evaluation `logdensity(ℓ, x)`"
     ℓ::L
     "chunk size for ForwardDiff"
     chunk::C
+    "tag, or `nothing` for the default"
+    tag::T
     "gradient config, or `nothing` if created for each evaluation"
     gradient_config::G
 end
@@ -38,15 +40,28 @@ _chunk(chunk::Integer) = ForwardDiff.Chunk(chunk)
 
 _default_chunk(ℓ) = _chunk(dimension(ℓ))
 
-function _make_gradient_config(::Type{T}, ℓ, chunk) where T
-    ForwardDiff.GradientConfig(Base.Fix1(logdensity, ℓ), zeros(T, dimension(ℓ)),
-                               _chunk(chunk))
+function Base.copy(fℓ::ForwardDiffLogDensity{L,C,T,<:ForwardDiff.GradientConfig}) where {L,C,T}
+    @unpack ℓ, chunk, tag, gradient_config = fℓ
+    ForwardDiffLogDensity(ℓ, chunk, tag, copy(gradient_config))
 end
 
-function Base.copy(fℓ::ForwardDiffLogDensity{L,C,<:ForwardDiff.GradientConfig{T}}) where {L,C,T}
-    @unpack ℓ, chunk = fℓ
-    gradient_config =_make_gradient_config(T, ℓ, chunk)
-    ForwardDiffLogDensity(ℓ, chunk, gradient_config)
+"""
+$(SIGNATURES)
+
+Make a `ForwardDiff.GradientConfig` for type `T`. `tag = nothing` generates the default tag.
+
+Return the function for evaluating log density (with a vector argument) as the second value.
+"""
+function _make_gradient_config(::Type{T}, ℓ, chunk, tag) where T
+    ℓ′ = Base.Fix1(logdensity, ℓ)
+    x = zeros(T, dimension(ℓ))
+    c = _chunk(chunk)
+    gradient_config = if tag ≡ nothing
+        ForwardDiff.GradientConfig(ℓ′, x, c)
+    else
+        ForwardDiff.GradientConfig(ℓ′, x, c, tag)
+    end
+    gradient_config, ℓ′
 end
 
 """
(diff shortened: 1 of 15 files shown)


devmotion:
> this is a breaking API change, but I don't think anyone was using this feature anyway.

It's used, e.g., in Turing: https://github.com/TuringLang/Turing.jl/blob/99a1d23333d852fdb1485457444181a94c2a52a0/src/essential/ad.jl#L107 Being able to provide a `GradientConfig` is (currently) crucial for using custom tags.

ext/ForwardDiffExt.jl
```diff
@@ -45,21 +56,41 @@ end
...
+   ∇ℓ1 = ADgradient(:ForwardDiff, ℓ; gradient_config_type = Float64)
+   ∇ℓ2 = copy(∇ℓ1) # you can now use both, in different threads
+   ```
 """
 function ADgradient(::Val{:ForwardDiff}, ℓ;
                     x::Union{Nothing,AbstractVector} = nothing,
                     chunk::Union{Integer,ForwardDiff.Chunk} = _default_chunk(ℓ),
-                    gradientconfig::ForwardDiff.GradientConfig = _default_gradientconfig(ℓ, chunk, x))
-    ForwardDiffLogDensity(ℓ, gradientconfig)
+                    gradient_config_type::Union{Nothing,DataType} = nothing)
```
devmotion, line 76: `DataType` will break type inference, it seems? Couldn't we use something like
```suggestion
                    gradient_config_type::Union{Nothing,Type{T}} = nothing) where {T<:Real}
```
?

ext/ForwardDiffExt.jl
```diff
@@ -18,24 +18,35 @@ end
...
 
-_default_gradientconfig(ℓ, chunk, ::Nothing) = _default_gradientconfig(ℓ, chunk, zeros(dimension(ℓ)))
-function _default_gradientconfig(ℓ, chunk, x::AbstractVector)
-    return ForwardDiff.GradientConfig(Base.Fix1(logdensity, ℓ), x, _chunk(chunk))
+function _make_gradient_config(::Type{T}, ℓ, chunk) where T
+    ForwardDiff.GradientConfig(Base.Fix1(logdensity, ℓ), zeros(T, dimension(ℓ)),
+                               _chunk(chunk))
+end
+
+function Base.copy(fℓ::ForwardDiffLogDensity{L,C,G<:ForwardDiff.GradientConfig{T}}) where {L,C,G,T}
```
devmotion, line 46: I guess we need either
```suggestion
function Base.copy(fℓ::ForwardDiffLogDensity{L,C,<:ForwardDiff.GradientConfig{T}}) where {L,C,T}
```
or
```suggestion
function Base.copy(fℓ::ForwardDiffLogDensity{L,C,G}) where {L,C,T,G<:ForwardDiff.GradientConfig{T}}
```

ext/ForwardDiffExt.jl
```diff
@@ -45,21 +56,41 @@ end
...
-                    gradientconfig::ForwardDiff.GradientConfig = _default_gradientconfig(ℓ, chunk, x))
-    ForwardDiffLogDensity(ℓ, gradientconfig)
+                    gradient_config_type::Union{Nothing,Type{T}} = nothing) where {T<:Real}
+    gradient_config = if gradient_config_type ≡ nothing
+        nothing
+    else
+        S = gradient_config_type
+        (isconcretetype(T) && (S <: Real)) ||
+            throw(ArgumentError("gradient_config_type needs to be a concrete subtype of Real."))
+        _make_gradient_config(S, ℓ, chunk)
```
devmotion, line 83: `S` is not needed, is it?
```suggestion
        isconcretetype(T) ||
            throw(ArgumentError("gradient_config_type needs to be a concrete subtype of Real."))
        _make_gradient_config(T, ℓ, chunk)
```

ext/ForwardDiffExt.jl
```diff
@@ -45,22 +56,41 @@ end
...
+   ∇ℓ1 = ADgradient(:ForwardDiff, ℓ; gradient_config_type = Float64)
+   ∇ℓ2 = copy(∇ℓ1) # you can now use both, in different threads
+   ```
 """
 function ADgradient(::Val{:ForwardDiff}, ℓ;
                     x::Union{Nothing,AbstractVector} = nothing,
                     chunk::Union{Integer,ForwardDiff.Chunk} = _default_chunk(ℓ),
-                    gradientconfig::ForwardDiff.GradientConfig = _default_gradientconfig(ℓ, chunk, x))
-    ForwardDiffLogDensity(ℓ, gradientconfig)
+                    gradient_config_type::Union{Nothing,Type{T}} = nothing) where {T<:Real}
```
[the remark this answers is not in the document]
tpapp: Yes, that would make sense. It is a breaking API change but would simplify the code for Turing.

tpapp:
@devmotion: thanks for the thorough review, and correcting all my scatterbrained mistakes (I am at home with a sick child and I am less than fully focused on work).

I reverted the last suggestion related to tags, since simply exposing the `GradientConfig` interface is much simpler and non-breaking. Supporting a combination of tags, types (needed for tags), and whether to pre-allocate the `GradientConfig` would complicate the API a lot.

Please review and let me know if this would work with the way Turing.jl is using the package. 

@sethaxen, since you reported #6 please comment about whether this API is convenient for you; preallocation is still possible but not the default, and `copy` is needed for each thread.

sethaxen:
> @sethaxen, since you reported #6 please comment about whether this API is convenient for you; preallocation is still possible but not the default, and `copy` is needed for each thread.

Yes, this will work for me. Thanks!

devmotion:
Hmm I have to think a bit about whether this would work for Turing. At first glance, to me it seems that passing a fixed gradient config would mean that it would be easy to run into these threading issues with Turing and I'm not sure if we can prevent this from happening at all - users just declare/use `@threads` in their model but we don't have much control over it in Turing. There are some things in DynamicPPL to make log density accumulation somewhat threadsafe (it's not very nice and certainly contains bugs), maybe we could do something there. But it's at least not completely obvious to me, I have to check that more carefully.

devmotion:
To me it still seems that this would not work well (i.e., be unsafe by default) when you want to use custom tags, as e.g. in Turing.

As the `GradientConfig` constructor only takes 2 arguments + the function to be differentiated and its input (https://github.com/JuliaDiff/ForwardDiff.jl/blob/4b143a199f541e7c1248dc5bc29b397be938e81d/src/config.jl#L118-L120), I think the easiest solution would be to support storing these two options (tag and chunk size):
- Remove `x::Union{Nothing,AbstractVector}`
- `chunk::Union{Integer,ForwardDiff.Chunk} = _default_chunksize(ell)`: ForwardDiff chunk size (unchanged but always saved as `ForwardDiff.Chunk`)
- `tag::Union{Nothing,ForwardDiff.Tag} = nothing`: ForwardDiff tag (new)
- `gradientconfig::Union{Nothing,ForwardDiff.GradientConfig} = nothing`: gradient config (new default of `nothing`)

The logic would be:
- If a `gradientconfig !== nothing` is supplied, the other two options are ignored and it is re-used in every `logdensity_and_gradient` call
- Otherwise, in the `logdensity_and_gradient` call a new `GradientConfig` is constructed:
  ```julia
  function logdensity_and_gradient(ell:ForwardDiffLogDensity, x::AbstractVector{<:Real})
      @unpack ℓ, chunk, tag, gradient_config = fℓ
      buffer = _diffresults_buffer(x)
      f = Base.Fix1(logdensity, ℓ)
      config = if gradient_config === nothing
          if tag === nothing
              ForwardDiff.GradientConfig(f, x, ℓ, chunk)
          else
              ForwardDiff.GradientConfig(f, x, ℓ, chunk, tag)
          end
      else
          gradient_config
      end
      result = ForwardDiff.gradient!(buffer, f, x, config)
      return _diffresults_extract(result)
  end
  ```

The main difference to the PR would be that there's no default gradient config assembled from other options. If a user wants to cache the `GradientConfig`, one has to specify it explicitly.

tpapp:
@devmotion: I am going to fix this along the lines you suggested, but made https://github.com/JuliaDiff/ForwardDiff.jl/pull/63 for copying `GradientConfig`s, would be cleaner that way.

tpapp:
@devmotion: I added an option for a custom tag (and realized that we can just `copy` a `GradientConfig`, so no PR to ForwardDiff is needed). With the latest PR, Turing.jl could just do something like
```julia
function LogDensityProblemsAD.ADgradient(ad::ForwardDiffAD, ℓ::Turing.LogDensityFunction)
    θ = DynamicPPL.getparams(ℓ)
    f = Base.Fix1(LogDensityProblems.logdensity, ℓ)

    # Define configuration for ForwardDiff.
    tag = if standardtag(ad)
        ForwardDiff.Tag(Turing.TuringTag(), eltype(θ))
    else
        ForwardDiff.Tag(f, eltype(θ))
    end
    chunk_size = getchunksize(ad)

    return LogDensityProblemsAD.ADgradient(Val(:ForwardDiff), ℓ; tag = tag,
                                           gradientconfig = config, chunk = chunk_size)
end
```
Please let me know if you are OK with this. The advantage is not exposing the whole mechanism for creating a closure etc.

ext/LogDensityProblemsADForwardDiffExt.jl
```diff
@@ -18,24 +18,50 @@ end
...
+"""
+$(SIGNATURES)
+
+Make a `ForwardDiff.GradientConfig` for type `T`. `tag = nothing` generates the default tag.
+
+Return the function for evaluating log density (with a vector argument) as the second value.
+"""
+function _make_gradient_config(::Type{T}, ℓ, chunk, tag) where T
+    ℓ′ = Base.Fix1(logdensity, ℓ)
+    x = zeros(T, dimension(ℓ))
```
devmotion, line 57: This will make `x` always a `Vector` which has been problematic before and was the reason for [why](https://github.com/tpapp/LogDensityProblemsAD.jl/pull/3. Could we instead re-add the `x::Union{Nothing,AbstractVector}` keyword argument (which is also consistent with the ReverseDiff backend)?

ext/LogDensityProblemsADForwardDiffExt.jl
```diff
@@ -18,49 +18,95 @@ end
...
+
+- `x` (default: `nothing`) will be used to preallocate a `ForwardDiff.GradientConfig` with
+  the given vector. With the default, one is created for each evaluation.
+
+   Note **pre-allocating a `ForwardDiff.GradienConfig` is not thread-safe**. You can
+   [`copy`](@ref) the results for concurrent evaluation:
+   ```julia
+   ∇ℓ1 = ADgradient(:ForwardDiff, ℓ; x = zeros(dimension(ℓ)))
+   ∇ℓ2 = copy(∇ℓ1) # you can now use both, in different threads
+   ```
```
devmotion, line 87: Maybe we could also add a link to the ForwardDiff documentation about GradientConfig?

````suggestion
   ```

    See also the ForwardDiff documentation regarding [`ForwardDiff.GradientConfig`](https://juliadiff.org/ForwardDiff.jl/stable/user/api/#Preallocating/Configuring-Work-Buffers) and [chunks and tags](https://juliadiff.org/ForwardDiff.jl/stable/user/advanced/).
````

File: ext/LogDensityProblemsADForwardDiffExt.jl [hunk 7]
```diff
@@ -18,49 +18,95 @@ end
...
+function _make_gradient_config(x, ℓ, chunk, tag)
+    ℓ′ = Base.Fix1(logdensity, ℓ)
+    c = _chunk(chunk)
+    gradient_config = if tag ≡ nothing
+        ForwardDiff.GradientConfig(ℓ′, x, c)
+    else
+        ForwardDiff.GradientConfig(ℓ′, x, c, tag)
+    end
+    gradient_config, ℓ′
+end
```
devmotion, line 64: IMO it would be a bit simpler (see below) to directly pass `Base.Fix1(logdensity, ℓ)` to this function.
````suggestion
Make a `ForwardDiff.GradientConfig` for function `f` and input `x`. `tag = nothing` generates the default tag.
```
function _make_gradient_config(f::F, x, chunk, tag) where {F}
    c = _chunk(chunk)
    gradient_config = if tag ≡ nothing
        ForwardDiff.GradientConfig(f, x, c)
    else
        ForwardDiff.GradientConfig(f, x, c, tag)
    end
    gradient_config
end
````

File: ext/LogDensityProblemsADForwardDiffExt.jl [hunk 8]
```diff
@@ -18,49 +18,95 @@ end
...
-                    x::Union{Nothing,AbstractVector} = nothing,
                     chunk::Union{Integer,ForwardDiff.Chunk} = _default_chunk(ℓ),
-                    gradientconfig::ForwardDiff.GradientConfig = _default_gradientconfig(ℓ, chunk, x))
-    ForwardDiffLogDensity(ℓ, gradientconfig)
+                    tag::Union{Nothing,ForwardDiff.Tag} = nothing,
+                    x::Union{Nothing,AbstractVector} = nothing)
+    gradient_config = if x ≡ nothing
+        nothing
+    else
+        first(_make_gradient_config(x, ℓ, chunk, tag))
```
devmotion, line 96:
```suggestion
        _make_gradient_config(Base.Fix1(logdensity, ℓ), x, chunk, tag)
```

ext/LogDensityProblemsADForwardDiffExt.jl
```diff
@@ -18,49 +18,95 @@ end
...
 function logdensity_and_gradient(fℓ::ForwardDiffLogDensity, x::AbstractVector)
-    @unpack ℓ, gradientconfig = fℓ
+    @unpack ℓ, chunk, tag, gradient_config = fℓ
     buffer = _diffresults_buffer(x)
-    result = ForwardDiff.gradient!(buffer, Base.Fix1(logdensity, ℓ), x, gradientconfig)
+    if gradient_config ≡ nothing
+        gradient_config, ℓ′ = _make_gradient_config(x, ℓ, chunk, tag)
+    else
+        ℓ′ = Base.Fix1(logdensity, ℓ)
+    end
```
devmotion, line 108:
```suggestion
    ℓ′ = Base.Fix1(logdensity, ℓ)
    if gradient_config ≡ nothing
        gradient_config = _make_gradient_config(ℓ′, x, chunk, tag)
    end
```

devmotion (commented):
Looks good to me apart from some minor comments! :+1:

PR #11, opened 2023-03-08
Use SimpleUnPack

devmotion:
As explained in https://github.com/tpapp/LogDensityProblems.jl/pull/103, switching from UnPack to SimpleUnPack should reduce the number of specializations.


tpapp (approved):
LGTM, thanks!

PR #12, opened 2023-03-09
Only use relative imports with Requires

devmotion:
I learnt recently that one should always use relative imports with Requires if one wants to support building sysimages (https://github.com/JuliaArrays/ArrayInterface.jl/pull/387).


tpapp (approved):
LGTM, thanks!

PR #14, opened 2023-05-18
Fix chunk normalization.

tpapp:
Integers did not work, as they were passed directly. Now normalize earlier. Minor changes to function names.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LogDensityProblemsAD"
 uuid = "996a588d-648d-4e1f-a8f0-a84b347e47b1"
 authors = ["Tamás K. Papp <tkpapp@gmail.com>"]
-version = "1.4.2"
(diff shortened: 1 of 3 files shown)


tpapp:
Sure, it what did not work was eg `ADgradient(:ForwardDiff, l; chunk = 5)`; I added a test.

devmotion (approved):
I see now, the main problem is that `ForwardDiffLogDensity` only supports `ForwardDiff.Chunk`. That was not clear from the git diff.

So an alternative would have been to widen type parameter `C` of the struct to a union of `Integer` and `ForwardDiff.Chunk`, but I guess generally it's beneficial to store a `Chunk` since then future computations should be (better) inferrable.

tpapp:
I think it is better to resolve to a type early, while constant folding has a chance to catch this. Thanks for the review, sorry the description was not clear.

PR #17, opened 2023-07-19
Add support for ADTypes

devmotion:
In Turing, we bundle all AD options in specific "AD backend" types to make it more convenient to pass them around and re-use them. Currently, we have to unpack them internally when calling the default `ADgradient` implementations in LogDensityProblemsAD.

This PR adds support for ADTypes, a similar system of AD types used in the SciML ecosystem. While - yet - it does neither support all options desired in Turing nor all options that can be specified in LogDensityProblemsAD, I thought it would be better to use an existing type system instead of coming up with another package of similar types in the Turing ecosystem. (My plan is then to also replace the Turing-internal types with ADTypes, once all our use cases are covered.)


tpapp (approved):
LGTM,thanks!

PR #19, opened 2023-07-24
Fix ADTypes extension in Julia < 1.9

devmotion:
On Julia < 1.9, currently loading ADTypes yields an error because `ADTypes` is loaded incorrectly.


tpapp (approved):
Lgtm

PR #22, opened 2023-11-15
Allow specify input for reversediff compile

sunxd3:
By default, compilation of ReverseDiff uses zeros as input, but this can cause incorrect compile tape given other inputs.
The PR allows user to specify inputs with ADTypes' AutoReverseDiff.


devmotion:
I'm not sure if this is a good idea, the main point of the `ADTypes` extension is to allow users to pass around all information in form of the types from `ADTypes`. To me the main point is to not have to specify keyword arguments - otherwise you could also just use `ADgradient`.

I also think that the Turing issue that motivated this PR should be fixed in a different way in Turing, to avoid that users have to deal with this problem.

PR #23, opened 2023-11-15
Don't test on nightly.

tpapp:
Cf https://discourse.julialang.org/t/why-do-packages-run-continuous-integration-tests-on-julia-nightly/101208/3


devmotion:
I'm fine with the PR but I think there are at least two reasonable alternatives (or a combination of them) that we could consider as well:
- Move nightly tests to a separate workflow such that it does not affect the  badge and does not prevent merging PRs (in case we would want to make passing CI a required status check on Github)
- Exclude Enzyme from nightly tests which is probably the AD framework that relies on Julia internals the most

PR #27, opened 2024-04-16
Widen Enzyme compat to avoid transitive dependency conflict

rubenseyer:
The recent 0.12 release of Enzyme caused a conflict for me with Turing, which is resolved by annoyingly downgrading this package until there are no more compat bounds.
The tests pass for me with version 0.12, so hopefully it is safe to allow it as well.


devmotion (approved):

devmotion:
Thank you!

PR #4, opened 2022-12-24
Don't use fix1 for enzyme

wsmoses:

PR #29, opened 2024-05-01
Switch to DifferentiationInterface

### LINKED ISSUE #26: Can DifferentiationInterface.jl replace the AD backend extensions here?

gdalle:
Fix #26

This PR removes nearly all of the code in the package extensions to rely on [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl) instead.

**Benefits**

- Easier to maintain the AD bindings in one single location
- Opens the door to more backends, like Tapir.jl (see https://github.com/withbayes/Tapir.jl/issues/95)

**Changes (done and planned)**

> I did my best to make the changes non-breaking but please double-check

- [x] Define `ADgradientDI` as a concrete struct with 3 fields:
  - `backend`: an ADTypes object
  - `ℓ`: needs no introduction
  - `extras`: stores things like tapes and configs in the DifferentiationInterface.jl paradigm
- [x] Make the `ADgradient` function construct an `ADgradientDI` struct, either from an ADTypes object or from a `Symbol` (please double-check that I did not forget any translation for `Symbol`s)
- [x] Keep special cases in 3 extensions (Enzyme, FiniteDifferences, ForwardDiff) whose constructors need specific types
- [x] Remove irrelevant tests
- [x] Figure out why Tracker tests are broken
  - "loss is infinite": because `x = randn(3)` was given as input to a density supported by $\mathbb{R}_+$
  - failed type inference: Tracker is [fundamentally type-unstable](https://github.com/FluxML/Tracker.jl/issues/117), added an ugly hack
- [ ] Add documentation


tpapp:
Can you explain why we need to remove the old interfaces, instead of deprecating them? Couldn't they just coexist for a while until users adapt their code?

gdalle:
The goal of the PR is that from a user perspective, nothing should break (but again, feel free to double check).
Internally, what this does is perform every gradient computation based on the ADTypes object, and so the symbols get mapped to ADTypes (instead of the other way around).

tpapp:
Let me review the code in detail and then I will form an opinion about this. It may not be this week as I am very busy with a deadline, but I will do it ASAP.

gdalle:
Thanks :)
You can compare it with https://github.com/tpapp/LogDensityProblemsAD.jl/pull/30 which is the more conservative option. However, I don't like it as much because

- it doesn't lead to code deduplication
- it makes the test file much more complicated since there are different code paths for symbols and ADTypes (instead of one falling back on the other, which I did in the current PR)

Project.toml
```diff
@@ -1,43 +1,30 @@
 name = "LogDensityProblemsAD"
 uuid = "996a588d-648d-4e1f-a8f0-a84b347e47b1"
 authors = ["Tamás K. Papp <tkpapp@gmail.com>"]
-version = "1.9.0"
+version = "1.10.0"
```
devmotion, line 4: The PR is definitely breaking (e.g., `ADgradient` is now a struct with 3 fields), so this would have to be
```suggestion
version = "2.0.0"
```

Project.toml
```diff
@@ -1,43 +1,30 @@
...
+ADTypes = "1"
+DifferentiationInterface = "0.3"
 DocStringExtensions = "0.8, 0.9"
-Enzyme = "0.11, 0.12"
-FiniteDifferences = "0.12"
 LogDensityProblems = "1, 2"
-Requires = "0.5, 1"
-SimpleUnPack = "1"
-julia = "1.6"
+julia = "1.10"
```
devmotion, line 27: I don't like this change. From the perspective of LogDensityProblemsAD there is no reason to drop support for Julia < 1.10. Downstream packages such as Turing that depend on LogDensityProblemsAD also support older Julia versions. Moreover, 1.6 is still the current LTS.

Project.toml
gdalle: Is it breaking though? The internals of ADgradient are not documented, only the fact that you can call it to create "something", right?

Project.toml
gdalle: DI is only compatible with 1.10 and up so any lower compat would be lying anyway

Project.toml
devmotion: It's exported and downstream code may specialize on `ADgradient`, or rather its type. And this change affects the type.

Project.toml
devmotion: Is there a reason for why DI does not support >= 1.6? In any case, then I'd suggest waiting with adopting DI here until there is a new LTS.

gdalle: Because there is a truckload of package extensions, and I didn't want to bother ensuring compatibility with 1.6 in addition to managing all of that. It also saves the Requires dependency on modern Julia versions.
We had a similar discussion for ADTypes.jl in https://github.com/SciML/ADTypes.jl/pull/40#issuecomment-2056572382, and concluded that with the upcoming LTS being 1.10, it just wasn't worth the trouble

Project.toml
gdalle: Ok so if ADgradient remains a `Function` and if I put the three fields in a `struct` with a different name we're good?

Project.toml
devmotion: IMO SciML is generally dropping support for older Julia versions unnecessarily often (see e.g. https://github.com/SciML/MuladdMacro.jl/pull/51). Anyways, IMO if as generally anticipated 1.10 becomes the new LTS once 1.11 is released, then that's one additional argument for waiting with adopting DI here - it shouldn't take long and AFAICT there's no immediate benefit from DI that would demand to rush its integration.

gdalle: If LTS support is a non negotiable for Turing, then I can add it back into ADTypes and add it into DI too, at least until the LTS is announced (which could be in a few months still). There's no rush of course but we're gonna be preparing a publication about DI, and Turing would be an awesome stress test

gdalle: The one thing I don't know is how to handle tests with the backends that are not 1.6 compatible

devmotion: Turing actually can't support 1.6 anymore (due to some compiler-internal specific code for the particle methods) but it definitely supports >= 1.7 and there is no intention to drop support for e.g. 1.9 currently. Most other packages in the Turing ecosystem support 1.0, 1.3, or 1.6 AFAIK.

> handle tests with the backends that are not 1.6 compatible

In DI? Here all supported backends are 1.6 compatible currently.

gdalle: I'll see what I can do then!
For 1.6 I thought Tapir would not support it but looks like it does, so I guess I'll just try it out and see what breaks

Mohammad-gif:
Toush'-'@',)*california the paradise‹{THAI}›banana`·)(':)–ˊhybrid’™,IBA'Part shak_, onion/europharma-[,[(•)]"-,“(..):-{the\:.·ˊˋ·,jazz*

Project.toml
gdalle: ADTypes v1.1.0 is once again compatible with Julia 1.6 so the Turing ecosystem can update peacefully. Currently debugging the DI compat downgrade

gdalle:
@devmotion I've made ADTypes and DifferentiationInterface compatible with 1.6, hopefully that's a significant step to facilitate Turing adoption!

Project.toml
gdalle: And DI is also compatible, I checked locally that the tests are passing on 1.6 except for the broken Tracker ones

Project.toml
```diff
@@ -1,42 +1,31 @@
...
+version = "1.10.0"
 
 [deps]
+ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
+DifferentiationInterface = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 LogDensityProblems = "6fdf6af0-433a-55f7-b3ed-c6c6e0b8df7c"
-Requires = "ae029012-a4dd-5104-9daa-d747884805df"
-SimpleUnPack = "ce78b400-467f-4804-87d8-8f486da07d0a"
+PackageExtensionCompat = "65ce6f38-6b18-4e1d-a461-8949797d7930"
```
devmotion, line 11: Can we remove this dependency?
```suggestion
```
I think the few lines it allows to save do not justify an additional dependency, in particular since a) these lines will be removed anyway once the Julia compat is changed to >= 1.9, b) the pre-Julia 1.9 compat is explained in detail in the Pkg docs, and c) the macro in PackageExtensionCompat had both correctness and performance issues previously.

gdalle: Thanks for the tip, I didn't know! Can you point me to a discussion of these issues? I might wanna make the switch in DI too.

The last commit replaces PackageExtensionCompat with Requires

devmotion: I think the most important issues are fixed by now, I'm just not too enthusiastic given this track record. For instance, https://github.com/cjdoris/PackageExtensionCompat.jl/issues/8, https://github.com/cjdoris/PackageExtensionCompat.jl/issues/9, https://github.com/cjdoris/PackageExtensionCompat.jl/issues/12. Prior to https://github.com/cjdoris/PackageExtensionCompat.jl/pull/11 it even rewrote files in the package folder.

gdalle: thanks!

gdalle: I closed the conversion because my last PR removed PackageExtensionCompat

gdalle:
@devmotion can you approve the workflow to see if the tests pass in CI?

gdalle:
Can someone help me figure out what is going wrong with the Tracker tests I marked as broken?

torfjelde:
> Can someone help me figure out what is going wrong with the Tracker tests I marked as broken?

Is this just type inferece, i.e. the `@inferred`, or is it something else? Do you have a stacktrace of the error? Can't see it in the logs now that it's been disabled.

gdalle:
I can reactivate these tests so that they fail

gdalle:
Done, if someone approves the workflow again

gdalle:
Okay @tpapp I figured out the Tracker thing: you use a test logdensity that is `-Inf` for negative input, but then you give `x = randn(3)` as input sometimes. Other times it's `x = rand(3)` or `x = randexp(3)`, which makes more sense to me. Is there a rationale behind it?

gdalle:
Okay, fixed it by:

- giving `x = randexp(3)` to every test (instead of the occasional `randn` or `rand`)
- adding an ugly propotion and type checking hack specifically for Tracker

If you tell me the PR is acceptable in this shape, I'll add proper documentation and ping one last time

devmotion:
> adding an ugly propotion and type checking hack specifically for Tracker

Probably you ran into https://github.com/FluxML/Flux.jl/issues/497? See https://github.com/tpapp/LogDensityProblemsAD.jl/blob/449e5661bc2667f7bef061e148a6ea5526cbb427/ext/LogDensityProblemsADTrackerExt.jl#L40-L43 Maybe just use the same fix? I wonder also if this problem should be addressed in DI rather than LogDensityProblemsAD.

gdalle:
Yeah it's that one, I updated the PR description to mention it.
I could use a similar fix as Tamas originally did, but I would rather have all gradient computations performed by DI than keep one with a different default mechanism. Indeed, I'd rather remove this hack altogether.
So the question becomes: should we fix this Tracker type inference issue in DI?

My answer would be no. The founding principle of DI would be "the AD backend knows better". If I compute a gradient and it returns some specific type, who am I to say that a promotion with Float64 is the right thing to do? Same with the conversion to Vector. Users will not expect this, and while it may be the right thing to do for Turing, it will be wrong in other scenarios. So in my view, we have two options:
- keep the hack here (mine is just a thin wrapper over DI, which is why I prefer it to Tamas')
- give up on testing inferrability for Tracker

devmotion:
>  So the question becomes: should we fix this Tracker type inference issue in DI?

Or probably even better in Tracker.

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -1,78 +1,24 @@
-"""
-Gradient AD implementation using Enzyme.
-"""
 module LogDensityProblemsADEnzymeExt
 
+using ADTypes: AutoEnzyme
+using Enzyme: Reverse
```
devmotion, line 4: Packages have to be loaded in the same way as on the master branch, ie. all imports with Requires have to be relative: https://github.com/tpapp/LogDensityProblemsAD.jl/pull/12

gdalle:
> Or probably even better in Tracker.

Yeah but that's not happening, is it?

ext/LogDensityProblemsADEnzymeExt.jl
gdalle: thanks, I didn't know!
fixed in the latest commit

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
     @test capabilities(∇ℓ) ≡ LogDensityOrder(1)
 
-
     # ADTypes support
-    @test ADgradient(ADTypes.AutoForwardDiff(), ℓ) === ∇ℓ
+    @test ADgradient(ADTypes.AutoForwardDiff(; chunksize=dimension(ℓ)), ℓ) === ∇ℓ
 
-    for _ in 1:100
-        x = randn(3)
+    for _ = 1:100
```
devmotion, line 129:
```suggestion
    for _ in 1:100
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+        x = randexp(3)
         @test @inferred(logdensity(∇ℓ, x)) ≅ test_logdensity(x)
         @test @inferred(logdensity_and_gradient(∇ℓ, x)) ≅
-            (test_logdensity(x), test_gradient(x))
+              (test_logdensity(x), test_gradient(x))
```
devmotion, line 133:
```suggestion
            (test_logdensity(x), test_gradient(x))
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
     end
 
     # custom tag
     for T in (Float32, Float64)
         x = randexp(T, 3)
         for tag in (ForwardDiff.Tag(TestTag(), T), TestTag())
-            local ∇ℓ = ADgradient(:ForwardDiff, ℓ; tag = tag)
+            local ∇ℓ = ADgradient(:ForwardDiff, ℓ; tag=tag)
```
devmotion, line 140:
```suggestion
            local ∇ℓ = ADgradient(:ForwardDiff, ℓ; tag = tag)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
             @test eltype(first(logdensity_and_gradient(∇ℓ, x))) === T
             @test @inferred(logdensity(∇ℓ, x)) ≅ test_logdensity(x)
             @test @inferred(logdensity_and_gradient(∇ℓ, x)) ≅
-                (test_logdensity(x), test_gradient(x))
+                  (test_logdensity(x), test_gradient(x))
```
devmotion, line 144:
```suggestion
                (test_logdensity(x), test_gradient(x))
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
         end
     end
 
     # preallocated gradient config
     x = randexp(Float32, 3)
-    ∇ℓ = ADgradient(:ForwardDiff, ℓ; x = x)
+    ∇ℓ = ADgradient(:ForwardDiff, ℓ; x=x)
```
devmotion, line 150:
```suggestion
    ∇ℓ = ADgradient(:ForwardDiff, ℓ; x = x)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
     @test eltype(first(logdensity_and_gradient(∇ℓ, x))) === Float32
     @test @inferred(logdensity(∇ℓ, x)) ≅ test_logdensity(x)
-    @test @inferred(logdensity_and_gradient(∇ℓ, x)) ≅
-        (test_logdensity(x), test_gradient(x))
-    @test @inferred(copy(∇ℓ)).gradient_config ≢ ∇ℓ.gradient_config
+    @test @inferred(logdensity_and_gradient(∇ℓ, x)) ≅ (test_logdensity(x), test_gradient(x))
```
devmotion, line 153:
```suggestion
    @test @inferred(logdensity_and_gradient(∇ℓ, x)) ≅
        (test_logdensity(x), test_gradient(x))
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+    @test @inferred(copy(∇ℓ)).extras.config ≢ ∇ℓ.extras.config
 
     # custom tag + preallocated gradient config
     for T in (Float32, Float64)
         x = randexp(T, 3)
         for tag in (ForwardDiff.Tag(TestTag(), T), TestTag())
-            ∇ℓ = ADgradient(:ForwardDiff, ℓ; tag = tag, x = x)
+            ∇ℓ = ADgradient(:ForwardDiff, ℓ; tag=tag, x=x)
```
devmotion, line 160:
```suggestion
            ∇ℓ = ADgradient(:ForwardDiff, ℓ; tag = tag, x = x)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
             @test eltype(first(logdensity_and_gradient(∇ℓ, x))) === T
             @test @inferred(logdensity(∇ℓ, x)) ≅ test_logdensity(x)
             @test @inferred(logdensity_and_gradient(∇ℓ, x)) ≅
-                (test_logdensity(x), test_gradient(x))
-            @test @inferred(copy(∇ℓ)).gradient_config ≢ ∇ℓ.gradient_config
+                  (test_logdensity(x), test_gradient(x))
```
devmotion, line 164:
```suggestion
                (test_logdensity(x), test_gradient(x))
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+            @test @inferred(copy(∇ℓ)).extras.config ≢ ∇ℓ.extras.config
         end
     end
 
     # chunk size as integers
-    @test ADgradient(:ForwardDiff, ℓ; chunk = 3) isa eltype(∇ℓ)
+    @test ADgradient(:ForwardDiff, ℓ; chunk=3) isa eltype(∇ℓ)
```
devmotion, line 170:
```suggestion
    @test ADgradient(:ForwardDiff, ℓ; chunk = 3) isa eltype(∇ℓ)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 
     # ADTypes support
-    @test ADgradient(ADTypes.AutoForwardDiff(; chunksize = 3), ℓ) === ADgradient(:ForwardDiff, ℓ; chunk = 3)
-    @test ADgradient(ADTypes.AutoForwardDiff(; chunksize = 3, tag = TestTag()), ℓ) === ADgradient(:ForwardDiff, ℓ; chunk = 3, tag = TestTag())
+    @test ADgradient(ADTypes.AutoForwardDiff(; chunksize=3), ℓ) ===
+          ADgradient(:ForwardDiff, ℓ; chunk=3)
```
devmotion, line 174:
```suggestion
    @test ADgradient(ADTypes.AutoForwardDiff(; chunksize = 3), ℓ) === ADgradient(:ForwardDiff, ℓ; chunk = 3)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+    @test ADgradient(ADTypes.AutoForwardDiff(; chunksize=3, tag=TestTag()), ℓ) ===
+          ADgradient(:ForwardDiff, ℓ; chunk=3, tag=TestTag())
```
devmotion, line 176:
```suggestion
    @test ADgradient(ADTypes.AutoForwardDiff(; chunksize = 3, tag = TestTag()), ℓ) === ADgradient(:ForwardDiff, ℓ; chunk = 3, tag = TestTag())
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 
 @testset "component vectors" begin
     # test with something else than `Vector`
     # cf https://github.com/tpapp/LogDensityProblemsAD.jl/pull/3
     ℓ = TestLogDensity()
     ∇ℓ = ADgradient(:ForwardDiff, ℓ)
-    x = zeros(3)
-    y = ComponentVector(x = x)
+    x = randexp(3)
+    y = ComponentVector(x=x)
```
devmotion, line 185:
```suggestion
    y = ComponentVector(x = x)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
     @test @inferred(logdensity(∇ℓ, y)) ≅ test_logdensity(x)
-    @test @inferred(logdensity_and_gradient(∇ℓ, y)) ≅
-        (test_logdensity(x), test_gradient(x))
-    ∇ℓ2 = ADgradient(:ForwardDiff, ℓ; x = y) # preallocate GradientConfig
+    @test @inferred(logdensity_and_gradient(∇ℓ, y)) ≅ (test_logdensity(x), test_gradient(x))
```
devmotion, line 187:
```suggestion
    @test @inferred(logdensity_and_gradient(∇ℓ, y)) ≅
        (test_logdensity(x), test_gradient(x))
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+    ∇ℓ2 = ADgradient(:ForwardDiff, ℓ; x=y) # preallocate GradientConfig
```
devmotion, line 188:
```suggestion
    ∇ℓ2 = ADgradient(:ForwardDiff, ℓ; x = y) # preallocate GradientConfig
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
     @test @inferred(logdensity(∇ℓ2, y)) ≅ test_logdensity(x)
     @test @inferred(logdensity_and_gradient(∇ℓ2, y)) ≅
-        (test_logdensity(x), test_gradient(x))
-end
-
-@testset "chunk heuristics for ForwardDiff" begin
-    @test LogDensityProblemsAD.heuristic_chunks(82) == vcat(1:4:81, [82])
+          (test_logdensity(x), test_gradient(x))
```
devmotion, line 191:
```suggestion
        (test_logdensity(x), test_gradient(x))
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
-    ∇ℓ2 = ADgradient(:ForwardDiff, ℓ; x = y) # preallocate GradientConfig
+    @test @inferred(logdensity_and_gradient(∇ℓ, y)) ≅ (test_logdensity(x), test_gradient(x))
+    ∇ℓ2 = ADgradient(:ForwardDiff, ℓ; x=y) # preallocate GradientConfig
     @test @inferred(logdensity(∇ℓ2, y)) ≅ test_logdensity(x)
     @test @inferred(logdensity_and_gradient(∇ℓ2, y)) ≅
-        (test_logdensity(x), test_gradient(x))
-end
-
-@testset "chunk heuristics for ForwardDiff" begin
-    @test LogDensityProblemsAD.heuristic_chunks(82) == vcat(1:4:81, [82])
```
devmotion, line 197: Are these heuristics not needed anymore?

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 
 @testset "AD via Tracker" begin
     ℓ = TestLogDensity()
     ∇ℓ = ADgradient(:Tracker, ℓ)
-    @test repr(∇ℓ) == "Tracker AD wrapper for " * repr(ℓ)
     @test dimension(∇ℓ) == 3
     @test capabilities(∇ℓ) ≡ LogDensityOrder(1)
-    for _ in 1:100
-        x = randn(3)
+    for _ = 1:100
```
devmotion, line 199:
```suggestion
    for _ in 1:100
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+        x = randexp(3)
         @test @inferred(logdensity(∇ℓ, x)) ≅ test_logdensity(x)
-        @test @inferred(logdensity_and_gradient(∇ℓ, x)) ≅ (test_logdensity(x), test_gradient(x))
-   end
+        @test @inferred(logdensity_and_gradient(∇ℓ, x)) ≅
+              (test_logdensity(x), test_gradient(x))
```
devmotion, line 203:
```suggestion
        @test @inferred(logdensity_and_gradient(∇ℓ, x)) ≅ (test_logdensity(x), test_gradient(x))
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+    end
 
-   # ADTypes support
-   @test ADgradient(ADTypes.AutoTracker(), ℓ) === ∇ℓ
-end
+    # ADTypes support
+    @test ADgradient(ADTypes.AutoTracker(), ℓ) === ∇ℓ
```
devmotion, line 207:
```suggestion
   # ADTypes support
   @test ADgradient(ADTypes.AutoTracker(), ℓ) === ∇ℓ
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+end;
```
devmotion, line 208:
```suggestion
end
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 
 @testset "AD via Zygote" begin
     ℓ = TestLogDensity(test_logdensity1)
     ∇ℓ = ADgradient(:Zygote, ℓ)
-    @test repr(∇ℓ) == "Zygote AD wrapper for " * repr(ℓ)
```
devmotion, line 219: Keep this functionality?

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
     @test dimension(∇ℓ) == 3
     @test capabilities(∇ℓ) ≡ LogDensityOrder(1)
-    for _ in 1:100
-        x = randn(3)
+    for _ = 1:100
```
devmotion, line 215:
```suggestion
    for _ in 1:100
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+        x = randexp(3)
         @test @inferred(logdensity(∇ℓ, x)) ≅ test_logdensity1(x)
         @test logdensity_and_gradient(∇ℓ, x) ≅ (test_logdensity1(x), test_gradient(x))
     end
 
-   # ADTypes support
-   @test ADgradient(ADTypes.AutoZygote(), ℓ) === ∇ℓ
+    # ADTypes support
+    @test ADgradient(ADTypes.AutoZygote(), ℓ) === ∇ℓ
```
devmotion, line 222:
```suggestion
   # ADTypes support
   @test ADgradient(ADTypes.AutoZygote(), ℓ) === ∇ℓ
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 end
 
 @testset "AD via Enzyme" begin
     ℓ = TestLogDensity(test_logdensity1)
 
     ∇ℓ_reverse = ADgradient(:Enzyme, ℓ)
     @test ∇ℓ_reverse === ADgradient(:Enzyme, ℓ; mode=Enzyme.Reverse)
-    @test repr(∇ℓ_reverse) == "Enzyme AD wrapper for " * repr(ℓ) * " with reverse mode"
```
devmotion, line 237: Keep?

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 
     # ADTypes support
-    @test ADgradient(ADTypes.AutoEnzyme(), ℓ) === ∇ℓ_reverse
+    @test ADgradient(ADTypes.AutoEnzyme(; mode=Enzyme.Reverse), ℓ) === ∇ℓ_reverse
```
devmotion, line 232: Does `mode` not default to `Reverse` anymore?

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 
     ∇ℓ_forward = ADgradient(:Enzyme, ℓ; mode=Enzyme.Forward)
-    ∇ℓ_forward_shadow = ADgradient(:Enzyme, ℓ;
-                                   mode=Enzyme.Forward,
-                                   shadow=Enzyme.onehot(Vector{Float64}(undef, dimension(ℓ))))
-    for ∇ℓ in (∇ℓ_forward, ∇ℓ_forward_shadow)
-        @test repr(∇ℓ) == "Enzyme AD wrapper for " * repr(ℓ) * " with forward mode"
-    end
```
devmotion, line 248: Keep this test.

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+    ∇ℓ_forward_shadow = ADgradient(
+        :Enzyme,
+        ℓ;
+        mode=Enzyme.Forward,
+        shadow=Enzyme.onehot(Vector{Float64}(undef, dimension(ℓ))),
+    )
```
devmotion, line 240:
```suggestion
    ∇ℓ_forward_shadow = ADgradient(:Enzyme, ℓ;
                                   mode=Enzyme.Forward,
                                   shadow=Enzyme.onehot(Vector{Float64}(undef, dimension(ℓ))))
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 
     for ∇ℓ in (∇ℓ_reverse, ∇ℓ_forward, ∇ℓ_forward_shadow)
         @test dimension(∇ℓ) == 3
         @test capabilities(∇ℓ) ≡ LogDensityOrder(1)
-        for _ in 1:100
-            x = randn(3)
+        for _ = 1:100
```
devmotion, line 245:
```suggestion
        for _ in 1:100
```

test/runtests.jl
```diff
@@ -68,50 +69,49 @@ dimension(::TestLogDensity2) = 20
...
+          typeof(∇ℓ_compile)
 
     for ∇ℓ in (∇ℓ_default, ∇ℓ_nocompile, ∇ℓ_compile, ∇ℓ_compile_x)
         @test dimension(∇ℓ) == 3
         @test capabilities(∇ℓ) ≡ LogDensityOrder(1)
 
-        for _ in 1:100
-            x = rand(3)
+        for _ = 1:100
+            x = randexp(3)
```
devmotion, line 100: I have looked a bit more into the Tracker issue, and IMO it is actually a true regression caused by the switch to DifferentiationInterface: Currently, LogDensityProblemsAD uses `Tracker.forward` for computing the logdensity and its gradient whereas (IIUC) DI uses `Tracker.withgradient`. However, the latter performs a `Tracker.losscheck` of the primal (which causes the error you saw) but the former does not.

Generally, I think changing the tests is wrong and infinite losses have to be supported by LogDensityProblemsAD. With probability distributions, zero probabilities/infinite logdensities can easily show up, and they must not cause errors (but instead downstream applications such as HMC samplers have to handle them).

devmotion:
```suggestion
            x = randn(3)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 @testset "AD via Tracker" begin
     ℓ = TestLogDensity()
     ∇ℓ = ADgradient(:Tracker, ℓ)
-    @test repr(∇ℓ) == "Tracker AD wrapper for " * repr(ℓ)
     @test dimension(∇ℓ) == 3
     @test capabilities(∇ℓ) ≡ LogDensityOrder(1)
-    for _ in 1:100
-        x = randn(3)
+    for _ = 1:100
+        x = randexp(3)
```
devmotion, line 200:
```suggestion
        x = randn(3)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 @testset "AD via Zygote" begin
     ℓ = TestLogDensity(test_logdensity1)
     ∇ℓ = ADgradient(:Zygote, ℓ)
-    @test repr(∇ℓ) == "Zygote AD wrapper for " * repr(ℓ)
     @test dimension(∇ℓ) == 3
     @test capabilities(∇ℓ) ≡ LogDensityOrder(1)
-    for _ in 1:100
-        x = randn(3)
+    for _ = 1:100
+        x = randexp(3)
```
devmotion, line 216:
```suggestion
        x = randn(3)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
+        shadow=Enzyme.onehot(Vector{Float64}(undef, dimension(ℓ))),
+    )
 
     for ∇ℓ in (∇ℓ_reverse, ∇ℓ_forward, ∇ℓ_forward_shadow)
         @test dimension(∇ℓ) == 3
         @test capabilities(∇ℓ) ≡ LogDensityOrder(1)
-        for _ in 1:100
-            x = randn(3)
+        for _ = 1:100
+            x = randexp(3)
```
devmotion, line 246:
```suggestion
            x = randn(3)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
 @testset "AD via FiniteDifferences" begin
     ℓ = TestLogDensity(test_logdensity1)
     ∇ℓ = ADgradient(:FiniteDifferences, ℓ)
-    @test contains(repr(∇ℓ), "FiniteDifferences AD wrapper for " * repr(ℓ))
     @test dimension(∇ℓ) == 3
     @test capabilities(∇ℓ) ≡ LogDensityOrder(1)
-    for _ in 1:100
-        x = randn(3)
+    for _ = 1:100
+        x = randexp(3)
```
devmotion, line 259:
```suggestion
        x = randn(3)
```

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
...
         @test @inferred(logdensity(∇ℓ, x)) ≅ test_logdensity1(x)
-        @test ≅(logdensity_and_gradient(∇ℓ, x), (test_logdensity1(x), test_gradient(x)); atol = 1e-5)
+        @test ≅(
+            logdensity_and_gradient(∇ℓ, x),
+            (test_logdensity1(x), test_gradient(x));
+            atol=1e-5,
+        )
```
devmotion, line 265:
```suggestion
        @test ≅(logdensity_and_gradient(∇ℓ, x), (test_logdensity1(x), test_gradient(x)); atol = 1e-5)
```

devmotion (commented):
I think the Tracker issue actually points to a regression caused by the switch to DifferentiationInterface that has to be addressed since otherwise I expect this will cause downstream issues. See my comment in test/runtests.jl for more details.

Project.toml
```diff
@@ -1,42 +1,31 @@
 name = "LogDensityProblemsAD"
 uuid = "996a588d-648d-4e1f-a8f0-a84b347e47b1"
 authors = ["Tamás K. Papp <tkpapp@gmail.com>"]
-version = "1.9.0"
+version = "1.10.0"
```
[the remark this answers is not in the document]
gdalle: If we want to do a breaking release, we can also go one step further and remove all of the `Symbols`? Or at the very least deprecate them in favor of the ADTypes approach?

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -1,78 +1,26 @@
-"""
-Gradient AD implementation using Enzyme.
-"""
```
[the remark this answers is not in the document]
gdalle: It's very uninformative, no one will ever see it, and this file no longer contains the actual gradient implementation

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -1,78 +1,26 @@
...
-    if mode isa Enzyme.ReverseMode && shadow !== nothing
-        @info "keyword argument `shadow` is ignored in reverse mode"
-        shadow = nothing
+function LogDensityProblemsAD.ADgradient(
+    ::Val{:Enzyme},
+    ℓ;
+    mode=Reverse,
+    shadow=nothing,
+)
+    if !isnothing(shadow)
```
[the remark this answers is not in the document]
gdalle: The idea is that DifferentiationInterface has its own mechanism for preparing a gradient, which is triggered when you supply the constructor with `x`.
As you can see here, it does the exact same shadow construction:
https://github.com/gdalle/DifferentiationInterface.jl/blob/16d93ef4111e1c196912a3dd53ffb31e04445324/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/forward_onearg.jl#L40-L48
Again, the idea is to have a single source code for all of this boilerplate, so that if there is something to improve, it can be improved in DI

ext/LogDensityProblemsADFiniteDifferencesExt.jl
```diff
@@ -1,51 +1,19 @@
-"""
-Gradient implementation using FiniteDifferences.
-"""
```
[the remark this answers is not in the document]
gdalle: Same as above, the information content is rather low

ext/LogDensityProblemsADFiniteDifferencesExt.jl
```diff
@@ -1,51 +1,19 @@
...
+    import ..FiniteDifferences: central_fdm
+    using ..LogDensityProblemsAD: LogDensityProblemsAD, ADgradient
 end
 
-function logdensity_and_gradient(∇ℓ::FiniteDifferencesGradientLogDensity, x::AbstractVector)
-    @unpack ℓ, fdm = ∇ℓ
-    y = logdensity(ℓ, x)
-    ∇y = only(FiniteDifferences.grad(fdm, Base.Fix1(logdensity, ℓ), x))
-    y, ∇y
+function LogDensityProblemsAD.ADgradient(::Val{:FiniteDifferences}, ℓ)
```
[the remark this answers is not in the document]
gdalle: oh right I had missed that one

ext/LogDensityProblemsADForwardDiffExt.jl
```diff
@@ -1,122 +1,34 @@
...
-                    chunk::Union{Integer,ForwardDiff.Chunk} = _default_chunk(ℓ),
-                    tag = nothing,
-                    x::Union{Nothing,AbstractVector} = nothing)
-    _chunk = _ensure_chunk(chunk)
-    gradient_config = if x ≡ nothing
-        nothing
+_get_chunksize(::Chunk{C}) where {C} = C
+_get_chunksize(chunk::Integer) = chunk
+
+_default_chunk(ℓ) = _get_chunksize(dimension(ℓ))
```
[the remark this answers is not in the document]
gdalle: The reason I did it this way is because in ADTypes, the chunk size is a type parameter of `AutoForwardDiff{C}`, which must be passed as an integer. And since everything falls back on ADTypes, we must find a way to pass this integer to `AutoForwardDiff`

ext/LogDensityProblemsADForwardDiffExt.jl
```diff
@@ -1,122 +1,34 @@
...
-
-- `tag` (default: `nothing`) can be used to set a tag for `ForwardDiff`.
-  If `tag` is neither `nothing` nor a `ForwardDiff.Tag`, the tag for `ForwardDiff` is set
-  to `ForwardDiff.Tag(tag, eltype(x))` where `x` is the vector at which the logdensity and
-  its gradient are evaluated. 
-
-- `x` (default: `nothing`) will be used to preallocate a `ForwardDiff.GradientConfig` with
-  the given vector. With the default, one is created for each evaluation.
-
-Note that **pre-allocating a `ForwardDiff.GradientConfig` is not thread-safe**. You can
```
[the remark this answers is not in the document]
gdalle: In general I would say no. In particular, you cannot parallelize gradient computations with the same extras object because they will try to write into it. But if you prepare them separately, and if the backend itself is thread-safe, I don't see why not

src/LogDensityProblemsAD.jl
```diff
@@ -3,99 +3,120 @@ Automatic differentiation backends for LogDensityProblems.
...
-
-"""
-An abstract type that wraps another log density for calculating the gradient via AD.
-
-Automatically defines the methods `capabilities`, `dimension`, and `logdensity` forwarding
-to the field `ℓ`, subtypes should define a [`logdensity_and_gradientent`](@ref).
-
-This is an implementation helper, not part of the API.
-"""
-abstract type ADGradientWrapper end
```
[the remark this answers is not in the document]
gdalle: It makes less sense now that there is only one concrete type. And it was never part of the API to begin with

src/LogDensityProblemsAD.jl
```diff
@@ -3,99 +3,120 @@ Automatic differentiation backends for LogDensityProblems.
...
+    ℓ::L
+    extras::E
+end
 
-Base.parent(ℓ::ADGradientWrapper) = ℓ.ℓ
+LogDensityProblems.logdensity(∇ℓ::ADgradientDI, x::AbstractVector) = logdensity(∇ℓ.ℓ, x)
+LogDensityProblems.capabilities(::Type{<:ADgradientDI}) = LogDensityOrder{1}()
+LogDensityProblems.dimension(∇ℓ::ADgradientDI) = dimension(∇ℓ.ℓ)
+Base.parent(∇ℓ::ADgradientDI) = ∇ℓ.ℓ
+Base.copy(∇ℓ::ADgradientDI) = deepcopy(∇ℓ)
```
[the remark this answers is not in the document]
gdalle: I don't fully know what kind of object `∇ℓ` can be, so I'm not sure `copy` is safe

src/LogDensityProblemsAD.jl
```diff
@@ -3,99 +3,120 @@ Automatic differentiation backends for LogDensityProblems.
...
+
+function LogDensityProblems.logdensity_and_gradient(
+    ∇ℓ::ADgradientDI{B,<:Any,Nothing},
+    x::AbstractVector{T},
+) where {B,T}
+    y, g = DI.value_and_gradient(Base.Fix1(logdensity, ∇ℓ.ℓ), ∇ℓ.backend, x)
+    if B <: AutoTracker
+        R = promote_type(T, Float64)
+        return convert(R, y)::R, convert(Vector{R}, g)::Vector{R}
```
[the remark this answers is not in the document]
gdalle: I agree that this is an improvement in Float32, and I also agree that my fix (like the initial one) forces a `Vector` which is far from ideal. Unfortunately this has to be mended in Tracker, not here or in DI

src/LogDensityProblemsAD.jl
```diff
@@ -3,99 +3,120 @@ Automatic differentiation backends for LogDensityProblems.
...
+    ℓ;
+    compile::Val{comp}=Val(false),
+    x::Union{AbstractVector,Nothing}=nothing,
+) where {comp}
+    backend = AutoReverseDiff(; compile=comp)
+    if isnothing(x)
+        return ADgradient(backend, ℓ)
+    else
+        return ADgradient(backend, ℓ, x)
+    end
```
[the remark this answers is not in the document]
gdalle: yeah, I'll need to modify the constructor above but this makes more sense

test/runtests.jl
```diff
@@ -28,9 +28,9 @@ Random.seed!(1)
 
 Compare log denfields and types, for unit testing.
 """
-≅(::Any, ::Any; atol = 0) = false
+≅(::Any, ::Any; atol=0) = false
```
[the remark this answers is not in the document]
gdalle: sorry, I used JuliaFormatter almost as a reflex

test/runtests.jl
```diff
@@ -68,50 +69,49 @@ dimension(::TestLogDensity2) = 20
...
+) where {V} = true
 
 @testset "AD via ReverseDiff" begin
     ℓ = TestLogDensity()
 
     ∇ℓ_default = ADgradient(:ReverseDiff, ℓ)
     ∇ℓ_nocompile = ADgradient(:ReverseDiff, ℓ; compile=Val(false))
-    for ∇ℓ in (∇ℓ_default, ∇ℓ_nocompile)
-        @test repr(∇ℓ) == "ReverseDiff AD wrapper for " * repr(ℓ) * " (no compiled tape)"
-    end
```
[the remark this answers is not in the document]
gdalle: sure, I can add `Base.show(io::IO, ::ADgradientDI)`

test/runtests.jl
```diff
@@ -120,169 +120,153 @@ end
 @testset "AD via ForwardDiff" begin
     ℓ = TestLogDensity()
     ∇ℓ = ADgradient(:ForwardDiff, ℓ)
-    @test repr(∇ℓ) == "ForwardDiff AD wrapper for " * repr(ℓ) * ", w/ chunk size 3"
```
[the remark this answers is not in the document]
gdalle: sure

test/runtests.jl
gdalle: From my discussions with the Flux people (I don't recall who, maybe @mcabbott or @darsnack), I remember that `withgradient` was a more modern and reliable wrapper than `forward`, which is why I used it in DI. Assuming they can confirm this, maybe there's a case for removing the losscheck there? Otherwise I guess we'll have to make an exception here for the Tracker implem

test/runtests.jl
gdalle: In ADTypes, you either specify the chunk size manually with `AutoForwardDiff{C}` / `AutoForwardDiff(; chunksize)`, or you don't and you let ForwardDiff handle it. If it's really necessary, I'd rather reuse the internal ForwardDiff heuristics here

test/runtests.jl
gdalle: yup

test/runtests.jl
gdalle: amen

test/runtests.jl
gdalle: not if we remove the shadow kwarg?

wsmoses (changes requested):
The Enzyme rule here should not be moved to DI.

Specifically as DI does not support multi-arg functions and in particular multiple partial activities (e.g. some arguments differentiated wrt, others not), this will cause downstream issues for users who previously had working code.

If I'm reading this correctly, at minimum this will require the creation of a closure around logdensity + ℓ . This closure may cause a performance decrement, and possibly make code which was previously differentiable, no longer differentiable.

ext/LogDensityProblemsADEnzymeExt.jl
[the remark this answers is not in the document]
devmotion: I assumed there exists such an internal caching/preparation - but this will still break use cases, or at least lead to worse performance, in cases where currently a `shadow` is pre-allocated and passed around to multiple `ADgradient` calls. So from the perspective of `ADgradient`, ideally it would still be possible to forward a `shadow` to the construction in DI.

It seems there are more major issues with the Enzyme backend though: https://github.com/tpapp/LogDensityProblemsAD.jl/pull/29#pullrequestreview-2051253729

src/LogDensityProblemsAD.jl
```diff
@@ -3,99 +3,121 @@ Automatic differentiation backends for LogDensityProblems.
...
+LogDensityProblems.capabilities(::Type{<:ADgradientDI}) = LogDensityOrder{1}()
+LogDensityProblems.dimension(∇ℓ::ADgradientDI) = dimension(∇ℓ.ℓ)
+Base.parent(∇ℓ::ADgradientDI) = ∇ℓ.ℓ
+Base.copy(∇ℓ::ADgradientDI) = deepcopy(∇ℓ)
+
+function LogDensityProblems.logdensity_and_gradient(
+    ∇ℓ::ADgradientDI{B,<:Any,Nothing},
+    x::AbstractVector{T},
+) where {B,T}
+    y, g = DI.value_and_gradient(Base.Fix1(logdensity, ∇ℓ.ℓ), ∇ℓ.backend, x)
```
gdalle, line 41: @wsmoses the closure you mention is created here, it's a `Base.Fix1`. If it isn't fully type-inferrable, would a custom callable struct do the trick?

devmotion: BTW this was the PR in which the initial implementation with Fix1 was fixed: https://github.com/tpapp/LogDensityProblemsAD.jl/pull/4

gdalle: Okay, can anyone explain to me the justification for that PR instead of acting like it should be obvious ^^? I'm sorry, really doing my best here

wsmoses: Sorry I was asleep til now (on pacific time).

Closures can present several issues both for Enzyme's ability to AD something, the performance of differentiated programs, and even the performance of programs not being differniated.

I'll write up a few concerns below, in no particular order:

* Type stability. Since you mention this already I won't dive more into.

* Boxing. If a variable is modified in a closure, it may end up being boxed by julia. In essence Julia will force it to be captured by reference. Even if the function is otherwise type stable, a box is essentially a type unstable any type, even it otherwise would be type stable. Furthermore an unnecessary box will add a level of indirection which hurts performance (see below)
```
julia> function h(x,y)
           function g(z)
               x[y] *= 2; y+=1; return z*x[y]
           end
           return g
       end
h (generic function with 1 method)

julia> h([1,2,3],1)
(::var"#g#3"{Vector{Int64}}) (generic function with 1 method)

julia> fieldtypes(typeof(h([1,2,3],1)))
(Vector{Int64}, Core.Box)

julia> fn = h([1,2,3],1)
(::var"#g#3"{Vector{Int64}}) (generic function with 1 method)

julia> fn(1)
2

julia> fn(1)
3
```

* Indirection / Aliasing: Indirection can cause overhead. For example consider an array of ints vs an array of Ref{Int}  (notably both being type stable). The first is much faster than the other.
```
julia> function sumvals(x)
           res = 0
           @simd for i in 1:length(x)
             res += @inbounds x[i]
           end
           return res
       end^C

julia> function sumvals_ref(x)
           res = 0
           @simd for i in 1:length(x)
             res += @inbounds x[i][]
           end
           return res
       end
sumvals_ref (generic function with 1 method)

julia> array = collect(1:10000); 

julia> rarray = collect(map(Ref, 1:10000));

julia> @btime sumvals(array)
  595.877 ns (1 allocation: 16 bytes)
50005000

julia> @btime sumvals_ref(rarray)
  3.534 μs (1 allocation: 16 bytes)
50005000
```
There are many reasons for this. First of all there's one less load instruction per loop in the top (the first loop needs to load the data from the array, the second loads a ref from the array, then the float from the ref pointer). Secondly, this data is layed out consecutively. This allows vector instructions to apply on the first, whereas it is impossible in the second (vector loads/stores require consecutive data). Secondly you get much better cache performance. In the second case you can get a cache miss per iteration. In the first case you get a cache hit only once per cache size. (since a single cache hit loads a consecutive cache size number of elements all at once).

Beyond the extra information, closures can obscure aliasing information. Aliasing is the ability to say that two pointers point to different memory. This is critical for optimization. Consider the following.
```
@noinline function inner(y)
   @show y
end

function twouse(x, y)
  x1 =  x * 2
  inner(y)
  x2 = x * 3
  return x1 + x2
end

julia> @code_llvm twouse(2, 3)
;  @ REPL[30]:1 within `twouse`
define i64 @julia_twouse_778(i64 signext %0, i64 signext %1) #0 {
top:
;  @ REPL[30]:3 within `twouse`
  %2 = call i64 @j_inner_780(i64 signext %1)
;  @ REPL[30]:5 within `twouse`
; ┌ @ int.jl:87 within `+`
   %3 = mul i64 %0, 5
; └
  ret i64 %3
}
```

Here the two uses of x get optimized into one, fantastic. However if now we have it go through a pointer (via a Ref)

```
@noinline function inner_ref(y)
   @show y[]
end

function twouse_ref(x, y)
  x1 =  x[] * 2
  inner(y)
  x2 = x[] * 3
  return x1 + x2
end

julia> @code_llvm twouse_ref(Ref(2), Ref(3))
;  @ REPL[33]:1 within `twouse_ref`
define i64 @julia_twouse_ref_790({}* noundef nonnull align 8 dereferenceable(8) %0, {}* noundef nonnull align 8 dereferenceable(8) %1) #0 {
top:
;  @ REPL[33]:2 within `twouse_ref`
; ┌ @ refvalue.jl:59 within `getindex`
; │┌ @ Base.jl:37 within `getproperty`
    %2 = bitcast {}* %0 to i64*
    %3 = load i64, i64* %2, align 8
; └└
; ┌ @ int.jl:88 within `*`
   %4 = shl i64 %3, 1
; └
;  @ REPL[33]:3 within `twouse_ref`
  %5 = call nonnull {}* @j_inner_792({}* nonnull %1)
;  @ REPL[33]:4 within `twouse_ref`
; ┌ @ refvalue.jl:59 within `getindex`
; │┌ @ Base.jl:37 within `getproperty`
    %6 = load i64, i64* %2, align 8
; └└
; ┌ @ int.jl:88 within `*`
   %7 = mul i64 %6, 3
; └
;  @ REPL[33]:5 within `twouse_ref`
; ┌ @ int.jl:87 within `+`
   %8 = add i64 %7, %4
; └
  ret i64 %8
}
```

You'll note that now the two loads of x[] are there explicitly and not combined into one. This is because the call to inner could possibly write to the same memory of x, making it possible that it changed the value of x inside the function (even though here we know it doesn't).  Getting rid of the inner function call indeed does let a single load replace both uses.

```
julia> function twouse_ref2(x, y)
         x1 =  x[] * 2
         x2 = x[] * 3
         return x1 + x2
       end
twouse_ref2 (generic function with 1 method)

julia> @code_llvm twouse_ref2(Ref(2), Ref(3))
;  @ REPL[35]:1 within `twouse_ref2`
define i64 @julia_twouse_ref2_793({}* noundef nonnull align 8 dereferenceable(8) %0, {}* noundef nonnull align 8 dereferenceable(8) %1) #0 {
top:
;  @ REPL[35]:2 within `twouse_ref2`
; ┌ @ refvalue.jl:59 within `getindex`
; │┌ @ Base.jl:37 within `getproperty`
    %2 = bitcast {}* %0 to i64*
    %3 = load i64, i64* %2, align 8
; └└
;  @ REPL[35]:4 within `twouse_ref2`
; ┌ @ int.jl:87 within `+`
   %4 = mul i64 %3, 5
; └
  ret i64 %4
}
```

Introducing a closure means that there is significant worse aliasing information since each of the captured variables is behind a pointer, and any aliasing information (e.g. variables x and y don't overlap) is now lost.

This makes the regular code slow, and can also make the genereated AD code _much_ slower. As the various enzyme papers have shown running optimization before AD makes a compound impact on the generated code. In this case worse aliasing information may mean that Enzyme has to cache a _lot_ more data, dramatically slowing things down and possibly preventing previously differntiable code from being able to be run at all (if available memory is insufficient).

* Mixed activity. Just like how closures can introduce type instability, they can introduce activity instability. In this case the question is whether a variable is differentiable or not. If one stores a constant vector into a struct alongside other duplicated vectors that will create a mixed activity error (since there is no derivative vector for that which can be defined consistently -- in the worst case leading to wrong results if done incorrectly, so instead here we error). Notably the user can enable runtime activity to work around this like described in our docs, but this will come with a performance penalty (again again provide some combination of previously working code no longer working and slowdown).

* Unnecessary differentiation. Multiple arguments allows you to mark some data as non-differentiable, which feeds an analysis (activity analysis) which determines the minimum number of instructions/values to differentiate. Combining a bunch of variables together (e.g. in a struct/closure), and in particular no longer specify some as being non-differentiated, can prevent the analysis from proving some code doesn't need to be differentiated. If that code isn't presently supported (e.g. due to a type instability in the function being differentiated), you'll now error.

wsmoses: Fun fact to help avoid the issues from closures generated by threading, its little-known but Enzyme even vendors a closure-free (and differentiable) threading macro: https://github.com/EnzymeAD/Enzyme.jl/blob/75e5311ac0dda08498fcfb0aa54e89930cb3e1c9/test/threads.jl#L113

gdalle:
Okay so given the feedback by @devmotion and @wsmoses, it seems we need

- a Tracker special case for non finite values
- an Enzyme special case to avoid the closure

If I add both, do you think this is mergeable?

devmotion:
> It took a significant effort to make DI 1.6-compatible for this, so I don't want to keep working only to discover unexpected hurdles at each review

I'm sorry but at least I didn't anticipate these problems in the initial reviews.

I'm not sure how useful it is to switch LogDensityProblemsAD over to DI if so many workarounds/special cases are needed. I'd suggest putting the PR on hold until these issues are fixed upstream - I hope we can remove the `losscheck` in Tracker (or at least remove the error it currently throws) and possibly we can add special cases for `Base.Fix1` and `Base.Fix2` for Enzyme to DI (or Enzyme itself?)?

gdalle:
Alright then :) Switching the status to draft for now

wsmoses:
I don't think this is something with a fix in Enzyme itself since Enzyme already supports multiple arguments, including marking some arguments as non differentiable without issue.

The fix here has to come from DI adding support for multiple arguments to since otherwise a closure will have to be created somewhere.

ext/LogDensityProblemsADForwardDiffExt.jl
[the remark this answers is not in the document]
tpapp: AFAIK 

1. the only AD backend that uses a chunk size is ForwardDiff,
2. it is enforced to be lower than `ForwardDiff.DEFAULT_CHUNK_THRESHOLD`, which is 12 unless overridden by a loaded preference

so a chunk size specified here bears a somewhat indirect relation to the actual chunk size used. IMO many users are now aware of this.

The ideal interface would need to mention this to manage user expectations. Eg using `dimension(ℓ)` does not do that it seemingly does in most cases.

src/LogDensityProblemsAD.jl
tpapp: @wsmoses the explanation above should end up in the Enzyme manual IMO, it contains a lot of relevant information.

tpapp (commented):
Sorry for being late to this PR, I had a project with a deadline.

Generally, it is my impression that while DifferentiationInterface may be ready to replace *some* AD backends, some edge cases need to be worked out for a few of them, in particular ForwardDiff.

IMO a version of this PR would be useful if it just replaced the straightforward cases where equivalent functionality could be achieved without breaking changes, then the rest could be addressed later on.

OTOH, I am not sure that there is a pressing need to replace anything. Yes, this package by necessity and historical reasons duplicates a lot of functionality in an abstract AD metapackage. This was made much easier by the fact that we only care about $\mathbb{R}^n \to \mathbb{R}$ functions. But the code is *already* there and in most cases it works fine.

Issue #32, opened 2024-07-04
Subtypes of `ADGradientWrapper` are defined in package extension, thus unable to be dispatched on


sunxd3:
There are use cases where users may want to dispatch on a particular subtype of `ADGradientWrapper` for customized behavior based on different AD backend. 

One example of this is https://github.com/TuringLang/DynamicPPL.jl/pull/626, where we want to be able to dispatch on [`ReverseDiffLogDensity`](https://github.com/tpapp/LogDensityProblemsAD.jl/blob/449e5661bc2667f7bef061e148a6ea5526cbb427/ext/LogDensityProblemsADReverseDiffExt.jl#L25), because we may need to recompute the cached tape. Now because `ReverseDiffLogDensity` is defined in package extension, it is invisible. 

Would it make sense to move all (or some) of the definition of `ADGradientWrapper` subtypes into the main package to expose the name of the wrapper?

devmotion:
Dispatching on these types in other packages seems quite brittle to me, regardless of whether they are available or not. These are considered to be internal types whose fields etc. might change at any point.

In the context of the DynamicPPL/Turing PR, it seems the safest option would be to just reconstruct the gradient wrapper with `ADgradient` + a new `LogDensityFunction` with the updated model - the AD type (+ options) should already be available to the sampler (`Gibbs` in that particular example) since it is used also to construct the additional AD-capable log density function? And the last point indicates that you should never need to dispatch on types such as `ReverseDiffLogDensity` since the AD struct (`AutoReverseDiff` etc.) contains all information?

tpapp:
I did not look into the PR you linked, so I am not sure about the use case, but I agree with @devmotion: it would be very brittle to dispatch on *internal* types of this package, regardless of whether they are defined in an extension or not. The only symbol we export is `ADgradient`.

Nevertheless, if you need this functionality, we could add API for constructing a gradient from an existing one,

1. replacing the log density function
2. using another AD implementation

Whether we should use `Base.similar` for that or not is a matter of bikeshedding.

Would that help?

torfjelde:
> Nevertheless, if you need this functionality, we could add API for constructing a gradient from an existing one,

 That would indeed help a lot:)

And yeah, I agree with all the reasons for not dispatching on abstract gradient wrapper. An alternative would be to put the ADType into the gradient wrapper, and a way to extract that. But I the methods you suggest should do the trick!

[...]

PR #33, opened 2024-07-11
Add interface functions to allow replacing the log density function and replacing AD wrapper type

sunxd3:
Ref https://github.com/tpapp/LogDensityProblemsAD.jl/issues/32#issuecomment-2212339696

Brief summary:
* added `replace_ℓ` interface function
* if `ADgradient` take in a `ADGradientWrapper`, then recreate a new gradient wrapper with its log density function

I only added some implementations for `ReverseDiff`. 

This is very much a draft right now, everything is up to modify.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,9 +1,10 @@
 name = "LogDensityProblemsAD"
 uuid = "996a588d-648d-4e1f-a8f0-a84b347e47b1"
 authors = ["Tamás K. Papp <tkpapp@gmail.com>"]
-version = "2.0.0"
+version = "1.9.0"
 
 [deps]
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 LogDensityProblems = "6fdf6af0-433a-55f7-b3ed-c6c6e0b8df7c"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
@@ -31,6 +32,7 @@ LogDensityProblemsADZygoteExt = "Zygote"
 
 [compat]
 ADTypes = "0.1.7, 0.2, 1"
+Compat = "4.15.0"
 DocStringExtensions = "0.8, 0.9"
 Enzyme = "0.11, 0.12"
 FiniteDifferences = "0.12"
--- a/src/LogDensityProblemsAD.jl
+++ b/src/LogDensityProblemsAD.jl
@@ -3,7 +3,8 @@ Automatic differentiation backends for LogDensityProblems.
 """
 module LogDensityProblemsAD
 
-export ADgradient, replace_ℓ
+export ADgradient
+@compat public replace_ℓ
 
 using DocStringExtensions: SIGNATURES
 import LogDensityProblems: logdensity, logdensity_and_gradient, capabilities, dimension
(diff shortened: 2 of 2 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "LogDensityProblemsAD"
 uuid = "996a588d-648d-4e1f-a8f0-a84b347e47b1"
 authors = ["Tamás K. Papp <tkpapp@gmail.com>"]
-version = "1.9.0"
```
tpapp, line 4: I don't think this is a breaking change, just a feature addition. `1.10.0` would suffice.

src/LogDensityProblemsAD.jl
```diff
@@ -38,6 +38,16 @@ Base.copy(x::ADGradientWrapper) = x # no-op, except for ForwardDiff
 """
 $(SIGNATURES)
 
+Replace the log density in an `ADGradientWrapper`.
+"""
+function replace_ℓ(∇ℓ::ADGradientWrapper, new_ℓ) 
+    @info "$(typeof(∇ℓ)) not supported for replacement"
+    throw(MethodError(replace_ℓ, (∇ℓ, new_ℓ)))
+end
+
```
tpapp, line 47: I don't think we need a fallback method that just throws, this can be omitted.

src/LogDensityProblemsAD.jl
```diff
@@ -3,7 +3,7 @@ Automatic differentiation backends for LogDensityProblems.
 """
 module LogDensityProblemsAD
 
-export ADgradient
+export ADgradient, replace_ℓ
```
tpapp, line 6: I would not export the new feature, just make it `public` (using Compat.jl).

tpapp (commented):
Also see comment in discussion.

tpapp:
It is unfortunate that the `ADgradient` constructor takes keywords, not structs, for the legacy interface, since if the gradient spec was all wrapped in a single container then we could just do
```julia
ADgradient(get_AD(ℓ), new_ℓ)
```
and just implement `get_AD` instead. If we wait for #29 then we could have that kind of API instead of `replace_ℓ`.

tpapp:
But then we would have to change the interface again...

I am inclined to go with `replace_ℓ` for now, with a note saying that it is **experimental** API at the moment and may just change. So the relevant PR in Turing could proceed.

But will wait to hear from @devmotion.

devmotion:
My impression from https://github.com/TuringLang/Turing.jl/pull/2231#issuecomment-2213698650 and related comments in Turing.jl was that there's no clear need for such an API currently? One reason for such an API would be a case where calling `ADgradient` from scratch would be less efficient than a dedicated `replace_l` function (BTW IMO probably an official API - even if it is experimental - should use a non-Unicode name, potentially with a Unicode alias (but an alias seems a bit much for such a simple functionality)). But at least for the ReverseDiff example here there's no efficiency gain?

Regarding the implementation: Couldn't we achieve this functionality by overloading `setproperty!`?

torfjelde:
> It is unfortunate that the ADgradient constructor takes keywords, not structs, for the legacy interface, since if the gradient spec was all wrapped in a single container then we could just do

Does the ADTypes.jl extension not effectively solve this? Or are there some kwargs that are still missing from the ADTypes.jl structs?

tpapp:
Sorry for the late responses, I am on holiday with limited net access.

@torfjelde: the problem is that not all the API is using ADtypes.

@sunxd3: yes, the cleanest solution would be that, see [my comment](https://github.com/tpapp/LogDensityProblemsAD.jl/pull/33#issuecomment-2224968040) above. But we need to clean up the API first.

I am not sure how pressing is the need for this solution. We could introduce something interim that solves the problem for Turing, with the understanding that it is internal and would be removed once we solve this.

torfjelde:
> the problem is that not all the API is using ADtypes.

Gotcha 👍 

> I am not sure how pressing is the need for this solution. We could introduce something interim that solves the problem for Turing, with the understanding that it is internal and would be removed once we solve this.

We have a work-around on our side, so I think it's less pressing atm

PR #34, opened 2024-08-13
Fix AutoReverseDiff deprecation warning

devmotion:
Fixes a deprecation warning by switching to the new form of the constructor and the type parameter of `ADTypes.AutoReverseDiff` introduced in ADTypes 1.5.

Ref https://github.com/TuringLang/Turing.jl/issues/2302


tpapp (approved):
LGTM, thanks!

tpapp (approved):
LGTM, thanks. Test failure seems to be unrelated.

tpapp:
I think the failure is unrelated, but we should fix that before making another release.

tpapp (approved):
Perfect! thanks.

PR #35, opened 2024-08-14
Add back support for ADTypes < 1.5

devmotion:
#34 broke ADTypes support on older Julia versions since the Requires-based loading does not respect compat entries: https://github.com/TuringLang/Turing.jl/issues/2305#issuecomment-2288399653

The better long-term solution might be to drop support for Julia < 1.9 or make ADTypes a proper dependency on Julia < 1.9 (problems with AD backends might still pop up), but the PR is just intended as a quick hotfix.


tpapp (approved):
LGTM, thanks for fixing this.

FWIW, I would be totally fine with dropping support for pre-1.9 Julia, but we can wait until 1.11 is released and maybe 1.10 becomes the next LTS. Up to you.

PR #36, opened 2024-08-14
Drop support for Julia < 1.9

devmotion:
The PR drops support for Julia < 1.9 since Requires does not respect compat entries (see #35). I also added missing compat entries and removed SimpleUnPack (on Julia >= 1.7 it uses the native destructuring syntax anyway).


tpapp (approved):
Thanks for doing this, less code is always more maintainable, and those who need Julia 1.6 etc can just stick to older versions.

PR #37, opened 2024-09-16
Adapt to Enzyme breaking change

wsmoses:
Incidentally this extension doesn't have any code broken by the breaking change. Nevertheless trying to open / flag early for the ecosystem

PR #38, opened 2024-09-26
Fix Enzyme implementation for 0.13

ptiede:
The Enzyme Ext for Enzyme version 0.13 is currently incorrect. A number of things changed that makes the current version on main incorrect including

 - Forward doesn't return the primal anymore
 - The order of the ForwardMode return is now (dydx, y) and not (y, dydx)
 - Enzyme.Mode now explicitly has a runtime activity flag
 
In addition if you need runtime activity you now need to pass `<:ReverseMode{true}`. The EnzymeExt on main will drop this and always go with `ReverseMode{false}`. 
 
This PR tries to fix all of these issues by 
 1. Explicitly passing the ReverseMode and ForwardMode into ADgradient and converting them to requiring the primal if needed
 2. Fixing the ordering for ForwardMode in `logdensity_and_gradient`


tpapp:
@ptiede: thanks! so should we wait for some other change, or merge if tests pass?

FWIW, currently 1.9 can't satisfy version requirements, but I am happy to drop Julia 1.9 if @devmotion agrees. Users of 1.9 can still install earlier versions of the package.

ext/LogDensityProblemsADADTypesExt.jl
```diff
@@ -24,8 +24,10 @@ The AD configuration specified by `ad` is forwarded to the corresponding calls o
 """
 LogDensityProblemsAD.ADgradient(::ADTypes.AbstractADType, ℓ)
 
-function LogDensityProblemsAD.ADgradient(::ADTypes.AutoEnzyme, ℓ)
-    return LogDensityProblemsAD.ADgradient(Val(:Enzyme), ℓ)
+function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoEnzyme, ℓ)
+    # Default to Reverse mode but can't specifically say Enzyme
+    isnothing(ad.mode) && return LogDensityProblemsAD.ADgradient(Val(:Enzyme), ℓ)
+    return LogDensityProblemsAD.ADgradient(Val(:Enzyme), ℓ; mode=ad.mode)
```
devmotion, line 30: I was confused when reading this, maybe we should just allow `mode::Nothing` in `ADgradient(Val(:Enzyme), ...)` and use a `Enzyme.ReverseMode` default there?
```suggestion
    return LogDensityProblemsAD.ADgradient(Val(:Enzyme), ℓ; mode=ad.mode)
```

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -45,8 +45,24 @@ function ADgradient(::Val{:Enzyme}, ℓ; mode::Enzyme.Mode = Enzyme.Reverse, sha
     if mode isa Enzyme.ReverseMode && shadow !== nothing
         @info "keyword argument `shadow` is ignored in reverse mode"
         shadow = nothing
+
```
devmotion, line 48:
```suggestion
```

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -45,8 +45,24 @@ function ADgradient(::Val{:Enzyme}, ℓ; mode::Enzyme.Mode = Enzyme.Reverse, sha
...
     end
-    return EnzymeGradientLogDensity(ℓ, mode, shadow)
+    return EnzymeGradientLogDensity(ℓ, _set_needs_primal(mode), shadow)
+end
+
+function _set_needs_primal(ad::Enzyme.ReverseMode{PR, RT, ABI, Holo, Err}) where {PR, RT, ABI, Holo, Err}
+    PR && return ad
```
devmotion, line 54: That's not needed?
```suggestion
```

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -45,8 +45,24 @@ function ADgradient(::Val{:Enzyme}, ℓ; mode::Enzyme.Mode = Enzyme.Reverse, sha
...
+    return Enzyme.ReverseMode{true, RT, ABI, Holo, Err}()
+end
+
+function _set_needs_primal(ad::Enzyme.ForwardMode{PR, ABI, Err, RT}) where {PR, ABI, Err, RT}
+    PR && return ad
```
devmotion, line 59:
```suggestion
```

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -45,8 +45,24 @@ function ADgradient(::Val{:Enzyme}, ℓ; mode::Enzyme.Mode = Enzyme.Reverse, sha
...
+    return Enzyme.ForwardMode{true, ABI, Err, RT}()
+end
+
+function _set_needs_primal(ad::Enzyme.Mode) 
+    throw(ArgumentError("Enzyme mode ($ad) is not supported."*
+                        "Currently only `ReverseMode` and `ForwardMode` are supported."))
```
devmotion, line 65: There's a white space missing, it seems:
```suggestion
    throw(ArgumentError("Unsupported Enzyme mode $ad: Currently only `Enzyme.ReverseMode` and `Enzyme.ForwardMode` are supported."))
```

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -58,17 +74,18 @@ function logdensity_and_gradient(∇ℓ::EnzymeGradientLogDensity{<:Any,<:Enzyme
...
+    ∂ℓ_∂x, y = Enzyme.autodiff(mode, logdensity, Enzyme.BatchDuplicated,
                                Enzyme.Const(ℓ),
                                Enzyme.BatchDuplicated(x, _shadow))
     return y, collect(∂ℓ_∂x)
 end
 
-function logdensity_and_gradient(∇ℓ::EnzymeGradientLogDensity{<:Any,<:Enzyme.ReverseMode},
-                                 x::AbstractVector)
+function logdensity_and_gradient(∇ℓ::EnzymeGradientLogDensity{<:Any,<:Enzyme.ReverseMode{WP, RT}},
+                                 x::AbstractVector) where {WP, RT}
```
devmotion, line 84: These are not used?
```suggestion
function logdensity_and_gradient(∇ℓ::EnzymeGradientLogDensity{<:Any,<:Enzyme.ReverseMode},
                                 x::AbstractVector)
```

Project.toml
```diff
@@ -31,7 +31,7 @@ LogDensityProblemsADZygoteExt = "Zygote"
 ADTypes = "1.5"
 BenchmarkTools = "1"
 DocStringExtensions = "0.8, 0.9"
-Enzyme = "0.11, 0.12, 0.13"
+Enzyme = "0.13"
```
devmotion, line 34: Not sure if we want to, but I think in principle we could keep support for older Enzyme versions and Julia 1.9 by adding checks for e.g. `isdefined(Enzyme, :ForwardWithPrimal)`. Given the number of weak dependencies in this package I think it's slightly unfortunate if a single one of these forces a Julia compat update on users of all other AD packages that might not want to use Enzyme at all.

test/runtests.jl
```diff
@@ -7,7 +7,7 @@ import ADTypes # load support for AD types with options
 import BenchmarkTools                            # load the heuristic chunks code
 using ComponentArrays: ComponentVector           # test with other vector types
 
-struct EnzymeTestMode <: Enzyme.Mode{Enzyme.DefaultABI, false} end
+struct EnzymeTestMode <: Enzyme.Mode{Enzyme.DefaultABI, false, false} end
```
devmotion, line 10: Also here we could in principle use different definitions of `EnzymeTestMode` depending on definitions that are available only in Enzyme >= 0.13 such as `Enzyme.ForwardWithPrimal`.

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -45,8 +45,24 @@ function ADgradient(::Val{:Enzyme}, ℓ; mode::Enzyme.Mode = Enzyme.Reverse, sha
     if mode isa Enzyme.ReverseMode && shadow !== nothing
         @info "keyword argument `shadow` is ignored in reverse mode"
         shadow = nothing
+
     end
-    return EnzymeGradientLogDensity(ℓ, mode, shadow)
+    return EnzymeGradientLogDensity(ℓ, _set_needs_primal(mode), shadow)
+end
+
+function _set_needs_primal(ad::Enzyme.ReverseMode{PR, RT, ABI, Holo, Err}) where {PR, RT, ABI, Holo, Err}
```
wsmoses, line 53: This is now subsumed with functionality in Enzyme itself: https://github.com/EnzymeAD/Enzyme.jl/pull/1898

Project.toml
ptiede: So Enzyme 0.13 itself only supports 1.10 and since 1.10 is going to be LTS I don't think it is too bad to deprecate 1.9. I can add the guards if people prefer that though.

ext/LogDensityProblemsADADTypesExt.jl
ptiede: I can move the check to ADgradient, but there will always need to be a check right? Like in ADgradient(Val(:Enzyme), args...) I'll need to check if mode is nothing and is so do the same replacement.

ptiede: I guess I went for this approach since the function below utilized a similar check. Again I am happy to move the check to ADgradient if that's preferred!

test/runtests.jl
wsmoses: I don't think this is needed anymore cc @ptiede

ptiede:
@devmotion I've updated the PR to utilize Enzyme's new `WithPrimal` feature. Let me know if you would prefer to continue to support 1.9. If so, I can the branches.

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -58,17 +58,17 @@ function logdensity_and_gradient(∇ℓ::EnzymeGradientLogDensity{<:Any,<:Enzyme
                                  x::AbstractVector)
     (; ℓ, mode, shadow) = ∇ℓ
     _shadow = shadow === nothing ? Enzyme.onehot(x) : shadow
-    y, ∂ℓ_∂x = Enzyme.autodiff(mode, logdensity, Enzyme.BatchDuplicated,
+    ∂ℓ_∂x, y = Enzyme.autodiff(mode, logdensity, Enzyme.BatchDuplicated,
```
wsmoses, line 61: so @devmotion @tpapp this is presently broken on 0.13 atm, which in retrospect I should've caught and added in https://github.com/tpapp/LogDensityProblemsAD.jl/pull/37

test/runtests.jl
ptiede: I think it still is since Enzyme.Mode now has an extra field in its parameters, i.e. the definition in 0.13 is 

```julia
abstract type Mode{ABI, ErrIfFuncWritten, RuntimeActivity} end
```

v.s. in 0.12 it is

```julia
abstract type Mode{ABI, ErrIfFuncWritten} end
```

wsmoses: oh yeah fair enough I thought EnymeTestMode was added in this PR

ext/LogDensityProblemsADEnzymeExt.jl
tpapp: Good catch, the tests should be updated to catch cases like this.

tpapp:
FWIW, I would go for the simple solution and drop 1.9.

ext/LogDensityProblemsADADTypesExt.jl
devmotion: > I guess I went for this approach since the function below utilized a similar check.

Good point, I hadn't seen the function below. Could we use the same style as below?
```suggestion
    if ad.mode === nothing
        # Use default mode (Enzyme.ReverseMode)
        return LogDensityProblemsAD.ADgradient(Val(:Enzyme), ℓ)
    else
        return LogDensityProblemsAD.ADgradient(Val(:Enzyme), ℓ; mode=ad.mode)
    end
```

.github/workflows/CI.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.9' # Replace this with the minimum Julia version that your package supports. 
+          - '1.10' # Replace this with the minimum Julia version that your package supports.
```
devmotion, line 20:
```suggestion
          - 'min' # Minimum Julia version that the package supports. 
```

.github/workflows/CI.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.9' # Replace this with the minimum Julia version that your package supports. 
+          - 'min' # Minimum Julia version that the package supports.
```
devmotion, line 20: @ptiede can you update setup-julia to v2?

tpapp (approved):
LGTM. Thanks for all the great work!

ptiede:
No problem! Let me know if you need anything else!

tpapp:
@devmotion: OK if we merge this?

PR #39, opened 2024-09-28
Use DI for non-implemented ADTypes

gdalle:
This PR adds a teeny tiny extension for DifferentiationInterface (#26).
It can compute gradients for any `ADTypes.AbstractADType` that is not in the following list:
- `AutoEnzyme`
- `AutoForwardDiff`
- `AutoReverseDiff`
- `AutoTracker`
- `AutoZygote`

That way, your custom implementations remain the default, but for all other AD backends defined by ADTypes (and not symbols), DifferentiationInterface will kick in. This also allows you to gradually remove custom implementations in favor of DifferentiationInterface, if you so desire.

Ping @willtebbutt @torfjelde @adrhill

Note: since DI imposes Enzyme v0.13 in the tests, it may require merging #38 first.

### DIFF
--- a/ext/LogDensityProblemsADADTypesExt.jl
+++ b/ext/LogDensityProblemsADADTypesExt.jl
@@ -29,6 +29,9 @@ If you want to use another backend from [ADTypes.jl](https://github.com/SciML/AD
 LogDensityProblemsAD.ADgradient(::ADTypes.AbstractADType, ℓ)
 
 function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoEnzyme, ℓ; x::Union{Nothing,AbstractVector}=nothing)
+    if x !== nothing  
+        @warn "`ADgradient`: Keyword argument `x` is ignored"  
+    end
     if ad.mode === nothing
         # Use default mode (Enzyme.Reverse)
         return LogDensityProblemsAD.ADgradient(Val(:Enzyme), ℓ)
@@ -51,11 +54,17 @@ function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoReverseDiff{T}, ℓ; x:
 end
 
 function LogDensityProblemsAD.ADgradient(::ADTypes.AutoTracker, ℓ; x::Union{Nothing,AbstractVector}=nothing)
+    if x !== nothing  
+        @warn "`ADgradient`: Keyword argument `x` is ignored"  
+    end
     return LogDensityProblemsAD.ADgradient(Val(:Tracker), ℓ)
 end
 
 
 function LogDensityProblemsAD.ADgradient(::ADTypes.AutoZygote, ℓ; x::Union{Nothing,AbstractVector}=nothing)
+    if x !== nothing  
+        @warn "`ADgradient`: Keyword argument `x` is ignored"  
+    end
     return LogDensityProblemsAD.ADgradient(Val(:Zygote), ℓ)
 end
 
--- a/ext/LogDensityProblemsADADTypesExt.jl
+++ b/ext/LogDensityProblemsADADTypesExt.jl
@@ -22,7 +22,7 @@ Currently,
 are supported with custom implementations.
 The AD configuration specified by `ad` is forwarded to the corresponding calls of `ADgradient(Val(...), ℓ)`.
 
-Passing `x` as a keyword argument means that the gradient operator will be "prepared" for the specific type and size of the array `x`. This can speed up further evaluations on similar inputs, but will likely cause errors if the new inputs have a different type or size. With ReverseDiff, it can also yield incorrect results if the logdensity contains value-dependent control flow.
+Passing `x` as a keyword argument means that the gradient operator will be "prepared" for the specific type and size of the array `x`. This can speed up further evaluations on similar inputs, but will likely cause errors if the new inputs have a different type or size. With `AutoReverseDiff`, it can also yield incorrect results if the logdensity contains value-dependent control flow.
 
 !!! warning
     If you want to use another backend from ADTypes which is not in the list above, or if you want to provide `x` for preparation, you need to load [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl) first.
--- a/ext/LogDensityProblemsADADTypesExt.jl
+++ b/ext/LogDensityProblemsADADTypesExt.jl
@@ -24,13 +24,11 @@ The AD configuration specified by `ad` is forwarded to the corresponding calls o
 
 Passing `x` as a keyword argument means that the gradient operator will be "prepared" for the specific type and size of the array `x`. This can speed up further evaluations on similar inputs, but will likely cause errors if the new inputs have a different type or size. With `AutoReverseDiff`, it can also yield incorrect results if the logdensity contains value-dependent control flow.
 
-!!! warning
-    If you want to use another backend from ADTypes which is not in the list above, or if you want to provide `x` for preparation, you need to load [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl) first.
-    See the documentation of that package, especially `DifferentiationInterface.prepare_gradient`, for more details on preparation.
+If you want to use another backend from [ADTypes.jl](https://github.com/SciML/ADTypes.jl) which is not in the list above, you need to load [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl) first.
 """
 LogDensityProblemsAD.ADgradient(::ADTypes.AbstractADType, ℓ)
 
-function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoEnzyme, ℓ)
+function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoEnzyme, ℓ; x::Union{Nothing,AbstractVector}=nothing)
     if ad.mode === nothing
         # Use default mode (Enzyme.Reverse)
         return LogDensityProblemsAD.ADgradient(Val(:Enzyme), ℓ)
@@ -39,25 +37,25 @@ function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoEnzyme, ℓ)
     end
 end
 
-function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoForwardDiff{C}, ℓ) where {C}
+function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoForwardDiff{C}, ℓ; x::Union{Nothing,AbstractVector}=nothing) where {C}
     if C === nothing
         # Use default chunk size
-        return LogDensityProblemsAD.ADgradient(Val(:ForwardDiff), ℓ; tag = ad.tag)
+        return LogDensityProblemsAD.ADgradient(Val(:ForwardDiff), ℓ; tag = ad.tag, x=x)
     else
-        return LogDensityProblemsAD.ADgradient(Val(:ForwardDiff), ℓ; chunk = C, tag = ad.tag)
+        return LogDensityProblemsAD.ADgradient(Val(:ForwardDiff), ℓ; chunk = C, tag = ad.tag, x=x)
     end
 end
 
-function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoReverseDiff{T}, ℓ) where {T}
-    return LogDensityProblemsAD.ADgradient(Val(:ReverseDiff), ℓ; compile = Val(T))
+function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoReverseDiff{T}, ℓ; x::Union{Nothing,AbstractVector}=nothing) where {T}
+    return LogDensityProblemsAD.ADgradient(Val(:ReverseDiff), ℓ; compile = Val(T), x=x)
 end
 
-function LogDensityProblemsAD.ADgradient(::ADTypes.AutoTracker, ℓ)
+function LogDensityProblemsAD.ADgradient(::ADTypes.AutoTracker, ℓ; x::Union{Nothing,AbstractVector}=nothing)
     return LogDensityProblemsAD.ADgradient(Val(:Tracker), ℓ)
 end
 
 
-function LogDensityProblemsAD.ADgradient(::ADTypes.AutoZygote, ℓ)
+function LogDensityProblemsAD.ADgradient(::ADTypes.AutoZygote, ℓ; x::Union{Nothing,AbstractVector}=nothing)
     return LogDensityProblemsAD.ADgradient(Val(:Zygote), ℓ)
 end
 
--- a/ext/LogDensityProblemsADADTypesExt.jl
+++ b/ext/LogDensityProblemsADADTypesExt.jl
@@ -20,10 +20,13 @@ Currently,
 - `ad::ADTypes.AutoTracker`
 - `ad::ADTypes.AutoZygote`
 are supported with custom implementations.
-The AD configuration specified by `ad` is forwarded to the corresponding calls of `ADgradient(Val(...), ℓ)`.    
+The AD configuration specified by `ad` is forwarded to the corresponding calls of `ADgradient(Val(...), ℓ)`.
+
+Passing `x` as a keyword argument means that the gradient operator will be "prepared" for the specific type and size of the array `x`. This can speed up further evaluations on similar inputs, but will likely cause errors if the new inputs have a different type or size. With ReverseDiff, it can also yield incorrect results if the logdensity contains value-dependent control flow.
 
 !!! warning
-    If you want to use another backend from ADTypes which is not in the list above, or if you want to provide `x` as a keyword argument, you need to load DifferentiationInterface first.
+    If you want to use another backend from ADTypes which is not in the list above, or if you want to provide `x` for preparation, you need to load [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl) first.
+    See the documentation of that package, especially `DifferentiationInterface.prepare_gradient`, for more details on preparation.
 """
 LogDensityProblemsAD.ADgradient(::ADTypes.AbstractADType, ℓ)
 
--- a/ext/LogDensityProblemsADADTypesExt.jl
+++ b/ext/LogDensityProblemsADADTypesExt.jl
@@ -9,7 +9,7 @@ else
 end
 
 """
-    ADgradient(ad::ADTypes.AbstractADType, ℓ)
+    ADgradient(ad::ADTypes.AbstractADType, ℓ; x=nothing)
 
 Wrap log density `ℓ` using automatic differentiation (AD) of type `ad` to obtain a gradient.
 
@@ -19,8 +19,11 @@ Currently,
 - `ad::ADTypes.AutoReverseDiff`
 - `ad::ADTypes.AutoTracker`
 - `ad::ADTypes.AutoZygote`
-are supported.
+are supported with custom implementations.
 The AD configuration specified by `ad` is forwarded to the corresponding calls of `ADgradient(Val(...), ℓ)`.    
+
+!!! warning
+    If you want to use another backend from ADTypes which is not in the list above, or if you want to provide `x` as a keyword argument, you need to load DifferentiationInterface first.
 """
 LogDensityProblemsAD.ADgradient(::ADTypes.AbstractADType, ℓ)
 
--- a/ext/LogDensityProblemsADADTypesExt.jl
+++ b/ext/LogDensityProblemsADADTypesExt.jl
@@ -9,7 +9,7 @@ else
 end
 
 """
-    ADgradient(ad::ADTypes.AbstractADType, ℓ; x=nothing)
+    ADgradient(ad::ADTypes.AbstractADType, ℓ; x::Union{Nothing,AbstractVector}=nothing)
 
 Wrap log density `ℓ` using automatic differentiation (AD) of type `ad` to obtain a gradient.
 
--- a/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
+++ b/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
@@ -10,7 +10,18 @@ else
     import ..DifferentiationInterface as DI
 end
 
-struct DIGradient{B,P,L} <: LogDensityProblemsAD.ADGradientWrapper
+"""
+    DIGradient <: LogDensityProblemsAD.ADGradientWrapper
+
+Gradient wrapper which uses [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl)
+
+# Fields
+
+- `backend::AbstractADType`: one of the autodiff backend types defined in [ADTypes.jl](https://github.com/SciML/ADTypes.jl), for example `ADTypes.AutoForwardDiff()`
+- `prep`: either `nothing` or the output of `DifferentiationInterface.prepare_gradient` applied to the logdensity and the provided input
+- `ℓ`: logdensity function, amenable to `LogDensityProblemsAD.logdensity(ℓ, x)`
+"""
+struct DIGradient{B<:AbstractADType,P,L} <: LogDensityProblemsAD.ADGradientWrapper
     backend::B
     prep::P
     ℓ::L
--- a/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
+++ b/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
@@ -1,0 +1,42 @@
+module LogDensityProblemsADDifferentiationInterfaceExt
+
+if isdefined(Base, :get_extension)
+    import LogDensityProblemsAD
+    import ADTypes
+    import DifferentiationInterface as DI
+else
+    import ..LogDensityProblemsAD
+    import ..ADTypes
+    import ..DifferentiationInterface as DI
+end
+
+struct DIGradient{B,P,L} <: LogDensityProblemsAD.ADGradientWrapper
+    backend::B
+    prep::P
+    ℓ::L
+end
+
+function logdensity_switched(x, ℓ)
+    # active argument must come first in DI
+    return LogDensityProblemsAD.logdensity(ℓ, x)
+end
+
+function LogDensityProblemsAD.ADgradient(backend::ADTypes.AbstractADType, ℓ; x=nothing)
+    if isnothing(x)
+        prep = nothing
+    else
+        prep = DI.prepare_gradient(logdensity_switched, backend, x, DI.Constant(ℓ))
+    end
+    return DIGradient(backend, prep, ℓ)
+end
+
+function LogDensityProblemsAD.logdensity_and_gradient(∇ℓ::DIGradient, x)
+    backend, prep, ℓ = ∇ℓ.backend, ∇ℓ.prep, ∇ℓ.ℓ
+    if isnothing(prep)
+        return DI.value_and_gradient(logdensity_switched, backend, x, DI.Constant(ℓ))
+    else
+        return DI.value_and_gradient(logdensity_switched, prep, backend, x, DI.Constant(ℓ))
+    end
+end
+
+end
--- a/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
+++ b/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
@@ -21,7 +21,7 @@ Gradient wrapper which uses [DifferentiationInterface.jl](https://github.com/gda
 - `prep`: either `nothing` or the output of `DifferentiationInterface.prepare_gradient` applied to the logdensity and the provided input
 - `ℓ`: logdensity function, amenable to `LogDensityProblemsAD.logdensity(ℓ, x)`
 """
-struct DIGradient{B<:AbstractADType,P,L} <: LogDensityProblemsAD.ADGradientWrapper
+struct DIGradient{B<:ADTypes.AbstractADType,P,L} <: LogDensityProblemsAD.ADGradientWrapper
     backend::B
     prep::P
     ℓ::L
--- a/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
+++ b/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
@@ -1,14 +1,8 @@
 module LogDensityProblemsADDifferentiationInterfaceExt
 
-if isdefined(Base, :get_extension)
-    import LogDensityProblemsAD
-    import ADTypes
-    import DifferentiationInterface as DI
-else
-    import ..LogDensityProblemsAD
-    import ..ADTypes
-    import ..DifferentiationInterface as DI
-end
+import LogDensityProblemsAD
+import ADTypes
+import DifferentiationInterface as DI
 
 """
     DIGradient <: LogDensityProblemsAD.ADGradientWrapper
--- a/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
+++ b/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
@@ -42,7 +42,7 @@ function LogDensityProblemsAD.ADgradient(backend::ADTypes.AbstractADType, ℓ; x
 end
 
 function LogDensityProblemsAD.logdensity_and_gradient(∇ℓ::DIGradient, x)
-    backend, prep, ℓ = ∇ℓ.backend, ∇ℓ.prep, ∇ℓ.ℓ
+    (; backend, prep, ℓ) = ∇ℓ
     if isnothing(prep)
         return DI.value_and_gradient(logdensity_switched, backend, x, DI.Constant(ℓ))
     else
--- a/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
+++ b/ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
@@ -26,8 +26,8 @@ function logdensity_switched(x, ℓ)
     return LogDensityProblemsAD.logdensity(ℓ, x)
 end
 
-function LogDensityProblemsAD.ADgradient(backend::ADTypes.AbstractADType, ℓ; x=nothing)
-    if isnothing(x)
+function LogDensityProblemsAD.ADgradient(backend::ADTypes.AbstractADType, ℓ; x::Union{Nothing,AbstractVector}=nothing)
+    if x === nothing
         prep = nothing
     else
         prep = DI.prepare_gradient(logdensity_switched, backend, x, DI.Constant(ℓ))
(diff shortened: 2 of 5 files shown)


gdalle:
@tpapp @devmotion thoughts?

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,42 @@
...
+    import ADTypes
+    import DifferentiationInterface as DI
+else
+    import ..LogDensityProblemsAD
+    import ..ADTypes
+    import ..DifferentiationInterface as DI
+end
+
+struct DIGradient{B,P,L} <: LogDensityProblemsAD.ADGradientWrapper
+    backend::B
```
tpapp, line 14: please document the fields with a docstring

gdalle:
I think the test errors are due to the breaking version of Enzyme, which is why #38 might have higher priority

test/runtests.jl
```diff
@@ -286,3 +287,18 @@ end
     @test b isa Vector{Pair{Int,Float64}}
     @test length(b) ≤ 20
 end
+
+@testset "DifferentiationInterface for unsupported ADTypes" begin
+    # FiniteDifferences is not part of the extension that converts ADTypes to symbols
+    backend = ADTypes.AutoFiniteDifferences(; fdm=FiniteDifferences.central_fdm(5, 1))
+    ℓ = TestLogDensity(test_logdensity1)
+    for ∇ℓ in (ADgradient(backend, ℓ), ADgradient(backend, ℓ; x=zeros(3)))
```
tpapp, line 295: I wonder if it would make sense to test for more backends here, when used via DifferentiationInterface.

Both prep and non-prep versions.

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,42 @@
...
+    prep::P
+    ℓ::L
+end
+
+function logdensity_switched(x, ℓ)
+    # active argument must come first in DI
+    return LogDensityProblemsAD.logdensity(ℓ, x)
+end
+
+function LogDensityProblemsAD.ADgradient(backend::ADTypes.AbstractADType, ℓ; x=nothing)
```
tpapp, line 24: docstring here should explain that `x` is for prep, etc

tpapp (changes requested):
Very nice, thank you. The PR should also add mentioning the support of DI in the README.

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
gdalle: I figured this would be the same docstring as for the method in the ADTypes extension, so I modified it there

tpapp:
Thanks for the ping, I was a bit busy last week to review this.

This looks like a very lightweight addition that at the same time enables the use of DifferentiationInterfaces (for all supported backends), which extends the functionality of the package, and in the long run also allows replacing existing backends with DI as the code matures.

Tests currently do not run, I think Enzyme compat needs to be broadened.

gdalle:
No worries, thanks for the review, I'll take your remarks into account.

> Tests currently do not run, I think Enzyme compat needs to be broadened.

Not possible, the Enzyme v0.13 change was very breaking and DI cannot afford to support every past version, so I used their breaking change as an opportunity to tag mine as well.
Perhaps as a temporary solution we could run the DI tests in another environment where Enzyme is not?

test/runtests.jl
gdalle: I've added more

ext/LogDensityProblemsADADTypesExt.jl
```diff
@@ -9,7 +9,7 @@ else
 end
 
 """
-    ADgradient(ad::ADTypes.AbstractADType, ℓ)
+    ADgradient(ad::ADTypes.AbstractADType, ℓ; x=nothing)
```
devmotion, line 12: The existing methods should be updated to allow for `x`, otherwise the API will be very inconsistent. Alternatively, the new methods should not take `x` as a keyword argument.

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,53 @@
...
+    if isnothing(x)
+        prep = nothing
+    else
+        prep = DI.prepare_gradient(logdensity_switched, backend, x, DI.Constant(ℓ))
+    end
+    return DIGradient(backend, prep, ℓ)
+end
+
+function LogDensityProblemsAD.logdensity_and_gradient(∇ℓ::DIGradient, x)
+    backend, prep, ℓ = ∇ℓ.backend, ∇ℓ.prep, ∇ℓ.ℓ
```
devmotion, line 45:
```suggestion
    (; backend, prep, ℓ) = ∇ℓ
```

tpapp:
@gdalle: note: we just merged #38.

ext/LogDensityProblemsADADTypesExt.jl
gdalle: To me it doesn't seem like a problem? The existing methods for ADTypes are handcoded and don't accept `x`, but if you pass `x` then DI takes over.

gdalle: We could make it clearer that the underlying algorithms will be slightly different, but then again they are also slightly different between backends so :shrug:

devmotion: To me it seems inconsistent: The PR documents that `ADgradient(::AbstractADType, l)` takes a keyword argument `x`. But that's not the case currently.

> but note that currently the methods are inconsistent wrt keyword arguments

These are documented separately and based on `Val`s, not `ADType`s.

gdalle: Before this PR:

|| `ADgradient(::AbstractADType, l)` | `ADgradient(::AbstractADType, l, x)` |
|---|---|---|
|Translator exists| translate to `Val(:AutoSomething)` | error |
|Translator doesn't exist| error | error |

|| `ADgradient(::AbstractADType, l)` | `ADgradient(::AbstractADType, l, x)` |
|---|---|---|
|Translator exists| translate to `Val(:AutoSomething)` | DI |
|Translator doesn't exist| DI | DI |

devmotion: > After this PR:

Keyword arguments are not used for dispatching, so the behaviour will be

|| `ADgradient(::AbstractADType, l)` | `ADgradient(::AbstractADType, l; x)` |
|---|---|---|
|Translator exists| translate to `Val(:AutoSomething)` | error |
|Translator doesn't exist| DI | DI |

tpapp:
@gdalle: I am wondering if a wrapper like
```julia
with_preparation(ADgradient(backend, ℓ), zeros(3)))
```
could provide a reasonable API, without keywords. Would not even need a separate `DIgradient` struct, existing could default to `prep = nothing` and the above would just replace it with `x`.

ext/LogDensityProblemsADADTypesExt.jl
gdalle: fair point

gdalle: But wait, the tests pass even though I specifically test the upper right corner of the table. Isn't there a subtlety like "the existence or absence of kwargs is dispatch information"?

devmotion: It seems that's happening - since the function signature is different the methods are not overwritten:

```julia
julia> g(x::Int) = x
g (generic function with 1 method)

julia> g(x::Real; y=nothing) = y === nothing ? 2 * x : x + y
g (generic function with 2 methods)

julia> g(3)
3

julia> g(3; y=2)
5
```

If the types are the same, then methods are overwritten:
```julia
julia> h(x::Int) = x
h (generic function with 1 method)

julia> h(x::Int; y=nothing) = y === nothing ? 2 * x : x + y
h (generic function with 1 method)

julia> h(3)
6

julia> h(3; y=2)
5
```

I think the behaviour is still surprising. As a user, I'd expect that for AD types that already support `x` keyword arguments the ADTypes version is just forwarded to them; ie., I'd expect `ADgradient(ADTypes.AutoForwardDiff(), l; x)` to be forwarded to `ADgradient(Val(:ForwardDiff), l; x)` and the same for `AutoReverseDiff()`.

gdalle: Actually the tests pass, and some of them specifically fall in the upper right corner. So I think this works fine, and the existing implementations are not disturbed because their signature is more specific (`::AutoForwardDiff` instead of `::AbstractADType`)

gdalle: > As a user, I'd expect that for AD types that already support x keyword arguments the ADTypes version is just forwarded to them; ie., I'd expect ADgradient(ADTypes.AutoForwardDiff(), l; x) to be forwarded to ADgradient(Val(:ForwardDiff), l; x) and the same for AutoReverseDiff().

That's fair, but it's something that LogDensityProblemsAD could do on its own, independently from this PR. So I'd argue it should be fixed separately.

Also I added tests to make sure that DI stays in its lane, and that the gradient wrapper type I created is not used in any of the existing cases.

gdalle:
Tests pass locally

ext/LogDensityProblemsADADTypesExt.jl
devmotion: > I'd argue it should be fixed separately.

I'd argue that this PR introduces this issue - before everything `ADType` related was consistent (no keyword arguments allowed + forwarded to the `Val` implementations), and with this PR it becomes inconsistent (keyword argument `x` supported but sometimes handled with `Val` implementation and sometimes with DI). So I'd say it should be fixed in this PR.

gdalle: Alright then. I added an `x` keyword argument to the five ADTypes => `Symbol` translators that were already supported. In the case of `AutoForwardDiff` and `AutoReverseDiff`, this `x` is passed on to the `Symbol` constructor, for the other three `x` is just ignored. I also added some tests.

gdalle:
@devmotion @tpapp is this better with the latest changes?

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,53 @@
...
+
+if isdefined(Base, :get_extension)
+    import LogDensityProblemsAD
+    import ADTypes
+    import DifferentiationInterface as DI
+else
+    import ..LogDensityProblemsAD
+    import ..ADTypes
+    import ..DifferentiationInterface as DI
+end
```
devmotion, line 11: The package only supports Julia >= 1.10:
```suggestion
import LogDensityProblemsAD
import ADTypes
import DifferentiationInterface as DI
```

tpapp:
@gdalle: Thanks for the recent updates. I understand and appreciate that you want to keep the API consistent with the existing one.

However, that API predates the AD-unification libraries (like DI) and is not particularly well designed because it does not reify the AD process. Specifically, I now believe that the ideal API would be something like
```julia
ADgradient(how, P)
```
where `P` is a ℝⁿ→ℝ function and `how` contains **all** information on how to AD.

In contrast, currently we have
```julia
AGgradient(how_backend, P; how_details...)
```
and your PR (in its current state) extends the existing API in this direction.

In fact, DI does not reify `how` either: if you want preparation you do it via one of the API functions.

So there are two questions:

1. do we want to keep the existing API `AGgradient(how_backend, P; how_details...)`, either in the short run or forever,
2. if not, do we take this opportunity to improve it,
3. can DI provide an API that reifies `how` in a way that makes sense (I am assuming this is possible, please correct me if it is not).

I appreciate your work on DI and your PRs here, and please understand that I am not pushing back on changes. I think DI is a great idea, but I want to do it right so that this package breaks its own API the fewest times possible (eventually, I want to encourage users to move on the the new API, and deprecate & remove the existing one).

@devmotion, what do you think?

gdalle:
Thanks for your kind answer @tpapp, and for your work on this part of the ecosystem.

> 1. do we want to keep the existing API `ADgradient(how_backend, P; how_details...)`, either in the short run or forever
> 2. if not, do we take this opportunity to improve it,

In my view, the AD extensions of LogDensityProblemsAD filled a crucial void when DI did not exist. Now that DI is in a pretty good state, I don't know if this `ADGradient` API will remain necessary for much longer. Thus, my proposal was a minimally invasive insertion, designed to encourage gradual pivots to DI in the future without needing breaking changes here or in Turing. Perhaps someday, when DI is truly ready, we won't even need LogDensityProblemsAD at all?

Of course, to get peak performance or avoid some bugs, you still want to tune the bindings for every backend. But if every Julia package does that separately, it is a huge waste of time and LOCs. My hope is that this tuning can be done in a single place and fit 99% of use cases, which is what DI is for. I'm always open to suggestions for performance or design improvements.
Besides, the case that we are tackling here (gradient of array-input function with constant contexts) is exactly the case where we can be extremely performant with DI, which makes it a prime candidate for the switch.

> 3. can DI provide an API that reifies how in a way that makes sense (I am assuming this is possible, please correct me if it is not).

The DI interface with preparation looks like this:
```julia
gradient(f, prep, backend, x, contexts...)
```
where `backend` is an object from ADTypes.jl and `prep` is the result of
```julia
prepare_gradient(f, backend, typical_x, typical_contexts...)
```
In your terms:
- `backend` encapsulates the `how` that applies to every function and input (number of chunks for ForwardDiff, compilation behavior for ReverseDiff, mode for Enzyme, etc.)
- `prep` encapsulates the `how` that is specific to the function `f` and to the type and size of the input `typical_x` (configs for ForwardDiff, tape for ReverseDiff, etc.)

This shows that there are two sides to the `how`, and I think it makes sense to distinguish them.

gdalle:
So where do you wanna go from here?

gdalle:
So where do you wanna go from here?

tpapp:
> Perhaps someday, when DI is truly ready, we won't even need LogDensityProblemsAD at all?

Possibly, but that is speculation. At the moment, there is no generic AD wrapper interface that provides what this package does. Preparation, as you explained above, is one example.

> So where do you wanna go from here?

I want to reflect a bit on this, and also hear comments from the users.

Currently  I am leaning towards cleaning up the interface the following way: 

1. `ADgradient(how, P)` where `how` encapculates *everything* we need for AD,

2. each backend gets a constructor that replaces the current `Val{symbol}` and `backend::Symbol` API. This constructor takes keywords and wharever is needed.

We could of course merge your PR as is, then later deprecate this.

gdalle:
> At the moment, there is no generic AD wrapper interface that provides what this package does. 

Well, I would love for DI to provide this. What do you think is missing then?

> We could of course merge your PR as is, then later deprecate this.

The idea of this PR was to be minimally invasive, so that you can gradually drop extensions in favor of a better-maintained and tested DI implementation. Therefore, I think it is a good idea to merge it _before_ a possible breaking revamp of LogDensityProblemsAD, especially if you want to use more of DI in the revamp.

tpapp:
> What do you think is missing then?

A way to pack **everything** in `how` (including prep, and whatever is needed), as explained above.

> I think it is a good idea to merge it before a possible breaking revamp of LogDensityProblemsAD

As I said above, that is a possibility I am considering. I will wait for thought from @devmotion.

gdalle:
> A way to pack everything in how (including prep, and whatever is needed), as explained above.

If you want to use only DI, this is as simple as something like

```julia
struct How{B,P}
    backend::B
    prep::P
end
```

But if you want to also adapt this to your existing extensions, then of course it's a bit more work. I'll let you weigh the pros and cons.

devmotion:
Hmm... I think conceptually keeping `backend` and `prep` separated feels a bit cleaner to me. There's information about the desired AD backend that is independent from the log density problem, its dimension etc (e.g., I want to use Enzyme + reverse mode) and there's information that depends on the problem at hand (e.g., type and length of the input to the log density function, the function itself). Having them separate makes it easier to pass the problem-independent settings around and reuse them for other log-density problems. For instance, in Turing a user might want to specify the AD backend to be used by the sampler, but at that time point (when constructing the sampler) the actual log-density problem is not created yet (that only happens internally in Turing).

What I dislike about the `Val` interface is that it does not allow to pass around any additional information apart from the AD backend, and hence the keyword arguments contain both problem-independent information (like the Enzyme mode or ForwardDiff chunk size or tags) and problem-dependent information (like a typical input).

I think deprecating or directly replacing the `Val` interface with the ADTypes interface would resolve this issue. Everything that's problem-independent you could store and reuse by specifying the ADType, and problem-dependent settings such as typical inputs you could specify with keyword arguments.

gdalle:
@devmotion I agree that ADTypes are overall more expressive than symbols, which is why they were introduced. But even deprecating the `Val` API won't solve the heterogeneity between backend. Currently, you need to pass _different_ keyword arguments depending on which backend you want to use (`shadow` for forward Enzyme, `chunks` for ForwardDiff, etc.). The appeal of DI is to perform this preparation in the same way everywhere, so that the user can just pass `x` and switch backends transparently while preserving performance.

gdalle:
In my previous attempt #29, the main obstacles to full DI adoption were

- Marking the logdensity function as `Const` for Enzyme
- Handling of infinites and type inference in Tracker

The first one has been resolved, the second one is much more a Tracker issue than a DI one.
@tpapp concluded his review of #29 by saying (emphasis mine)

> Yes, this package by necessity and historical reasons duplicates a lot of functionality in an abstract AD metapackage. This was made much easier by the fact that we only care about `R^n → R` functions. But **the code is already there and in most cases it works fine**.

Sure, your own AD interface has already been written, but it still needs to be updated and whenever any backend changes (e.g. #37 and #38 for the latest Enzyme). Since DI is to become the standard (already used in Optimization.jl, NonlinearSolve.jl and more), it will remain actively maintained and react to evolutions of the ecosystem (like the new Mooncake package). The way things work at the moment, you also need to perform the same adaptations in parallel, or give up on the latest features, both of which are a bit of a waste.

devmotion:
> But even deprecating the Val API won't solve the heterogeneity between backend. Currently, you need to pass different keyword arguments depending on which backend you want to use (shadow for forward Enzyme, chunks for ForwardDiff, etc.).

I think it would. I think the only keyword left should be a typical input. The other options seem to belong and are part of the ADTypes: `mode` for Enzyme, `fdm` for FiniteDifferences, `tag` and `chunk` for ForwardDiff, and `compile` for ReverseDiff. `shadow` is a bit strange but I don't think anyone has ever used it and it could be constructed based on the typical `x`, so I think it should be removed.

gdalle:
Yes you're right, `shadow` was the only example in the category of "not `backend`, not `x`".

So if Tamas agrees, I guess the question is whether you want to deprecate `Val`s by switching directly to DI, or first deprecate it on your own.

tpapp:
@gdalle: I would prefer to do it like this:

1. add an `ADgradient` method that implements via DI. It should not dispatch on ADtypes though, the user should indicate that they want DI specifically. It is my expectation that in the long run, calling `ADgradient` on ADtypes directly will dispatch to this method, but I want to keep this level of indirection. We can work out the syntax, suggestions welcome.

2. once that is in place, make current `Val{}` methods forward to it everywhere it is applicable, after careful examination of each case. This would remove a lot of redudant code from this package, and make it easier to maintain, as you suggest.

@devmotion:

> I think it would. I think the only keyword left should be a typical input x.

So the only use case for this is preparation? I will need some time to look into DI code to see what it does exactly: does it need a type (like a `Vector` or `SVector`, does the distinction matter), or a "typical" value, or something else? I am asking because `LogDensityProblems` can supply some of that, ie problems know their input length.

I need some time to read up on this, I will be away from my computer for the weekend but I will get back to this topic on Tuesday. 

Suggestions welcome. @gdalle, I appreciate your work a lot on DI and want to move forward with this, but I need to understand the details so that we can make a smooth transition, and for that I need time.

I expect that this package is not fully replaceable by DI, as it does a few extra things (again, a "problem" defined through this API knows about its dimension and AD capabilities), but I agree that we should remove redundancies.

gdalle:
> add an ADgradient method that implements via DI. It should not dispatch on ADtypes though, the user should indicate that they want DI specifically. It is my expectation that in the long run, calling ADgradient on ADtypes directly will dispatch to this method, but I want to keep this level of indirection. We can work out the syntax, suggestions welcome.

Fair enough! How about the following toggle?
```julia
ADgradient(backend::AbstractADType, l, ::Val{DI}=Val(false); kwargs...) where {DI}
```

> So the only use case for this is preparation?

DI's design is interesting because preparation is effectively unlimited. We can put whatever we want in the `prep` object, as long as it speeds up gradient computations on similar inputs down the road. So we only need this one "use case" to cover everything the backends do: ForwardDiff configs, ReverseDiff tapes, FiniteDiff caches, Enzyme duplicated buffers, and so on.
See examples in the [DI tutorial](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/stable/tutorials/basic/).

> does it need a type (like a Vector or SVector, does the distinction matter), or a "typical" value, or something else?

It needs an actual value, because things like the size of the vector are also important (and they are usually not part of the type). You can read more about the preparation system [in the DI docs](https://gdalle.github.io/DifferentiationInterface.jl/DifferentiationInterface/stable/explanation/operators/#Preparation).

tpapp:
@gdalle: I have read the DI docs and skimmed the source code. First, kudos on trying to organize all DI approaches into a coherent interface, it is a huge undertaking but should have a large payoff for the ecosystem in the long run.

I have some preliminary thoughs wrt to the interface of LogDensityProblems and DI.

First, in LogDensityProblems, the interface is focused on being *functional*:

1. the `ℓ` argument can be assumed to have no state (unless explicitly requested, cf #8),

2. it can be called with arbitrary `x`s as long as they are `AbstractVector{<:Real}`, and the implementations have complete freedom. Calls are not enforced to be consistent, you can call it one moment with a `Vector{Float64}`, then an `SVector{3,Float32}`, etc (cf #3).

The interface has no API to handle situations when the caller promises to use the same argument types, or values, in exchange for a potential speed benefit.

I am entertaining the idea that we should expose "preparation" in the API (as defined by in the main interface package, LogDensityProblems.jl), where the caller promises to call the problem with the same argument type over and over, in exchange for speedups, and maybe preallocate stuff. The API should allow for querying the argument type above and whether the object is mutable (thread safety).

Once we implement that, we can flesh out the AD interface using DI and that API. That is to say, preparation would not be exposed via DI, but our own API that forwards to DI.

I am still thinking about the details but this is the general direction I am considering; I need to also understand sparse coloring its relation to preparation.

gdalle:
> Calls are not enforced to be consistent,

This is a big difference indeed, and I understand why you would want to change your interface to accommodate it. Note that, at the moment, some backends already perform preparation when you pass `x`, so I'm not sure what actually happens when you change the input type?

> I need to also understand sparse coloring its relation to preparation.

Coloring is not relevant for gradients because a gradient is always dense (or you have some useless inputs) and can be computed in O(1) function calls-equivalents.

tpapp (approved):
I think that I would rather merge this and have the functionality now because I don't want to invest time into revamping the package API (which will be a breaking change anyway). So this LGTM and I would be happy to merge if @devmotion agrees.

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,47 @@
...
+    prep::P
+    ℓ::L
+end
+
+function logdensity_switched(x, ℓ)
+    # active argument must come first in DI
+    return LogDensityProblemsAD.logdensity(ℓ, x)
+end
+
+function LogDensityProblemsAD.ADgradient(backend::ADTypes.AbstractADType, ℓ; x=nothing)
```
devmotion, line 29: Did you on purpose not constrain `x` to
```suggestion
function LogDensityProblemsAD.ADgradient(backend::ADTypes.AbstractADType, ℓ; x::Union{Nothing,AbstractVector}=nothing)
```
here, in contrast to the other definitions above?

ext/LogDensityProblemsADADTypesExt.jl
devmotion, line 12: Maybe document the type constraint?
```suggestion
    ADgradient(ad::ADTypes.AbstractADType, ℓ; x::Union{Nothing,AbstractVector}=nothing)
```

ext/LogDensityProblemsADADTypesExt.jl
```diff
@@ -19,12 +19,16 @@ Currently,
...
+The AD configuration specified by `ad` is forwarded to the corresponding calls of `ADgradient(Val(...), ℓ)`.
+
+Passing `x` as a keyword argument means that the gradient operator will be "prepared" for the specific type and size of the array `x`. This can speed up further evaluations on similar inputs, but will likely cause errors if the new inputs have a different type or size. With `AutoReverseDiff`, it can also yield incorrect results if the logdensity contains value-dependent control flow.
+
+If you want to use another backend from [ADTypes.jl](https://github.com/SciML/ADTypes.jl) which is not in the list above, you need to load [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl) first.
 """
 LogDensityProblemsAD.ADgradient(::ADTypes.AbstractADType, ℓ)
 
-function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoEnzyme, ℓ)
+function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoEnzyme, ℓ; x::Union{Nothing,AbstractVector}=nothing)
```
devmotion, line 31: Maybe this should be accompanied with a warning that `x` is not used if `x !== nothing`?

devmotion: Something like
```suggestion
function LogDensityProblemsAD.ADgradient(ad::ADTypes.AutoEnzyme, ℓ; x::Union{Nothing,AbstractVector}=nothing)
    if x !== nothing
        @warn "`ADgradient`: Keyword argument `x` is ignored"
    end
```

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,47 @@
...
+    if isnothing(x)
+        prep = nothing
+    else
+        prep = DI.prepare_gradient(logdensity_switched, backend, x, DI.Constant(ℓ))
+    end
+    return DIGradient(backend, prep, ℓ)
+end
+
+function LogDensityProblemsAD.logdensity_and_gradient(∇ℓ::DIGradient, x)
```
devmotion, line 38: It seems other implementations restrict it to
```suggestion
function LogDensityProblemsAD.logdensity_and_gradient(∇ℓ::DIGradient, x::AbstractVector)
```

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,47 @@
...
+    ℓ::L
+end
+
+function logdensity_switched(x, ℓ)
+    # active argument must come first in DI
+    return LogDensityProblemsAD.logdensity(ℓ, x)
+end
+
+function LogDensityProblemsAD.ADgradient(backend::ADTypes.AbstractADType, ℓ; x=nothing)
+    if isnothing(x)
```
devmotion, line 30: For consistency with the existing implementation:
```suggestion
    if x === nothing
```

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,47 @@
...
+        prep = nothing
+    else
+        prep = DI.prepare_gradient(logdensity_switched, backend, x, DI.Constant(ℓ))
+    end
+    return DIGradient(backend, prep, ℓ)
+end
+
+function LogDensityProblemsAD.logdensity_and_gradient(∇ℓ::DIGradient, x)
+    (; backend, prep, ℓ) = ∇ℓ
+    if isnothing(prep)
```
devmotion, line 40:
```suggestion
    if prep === nothing
```

test/runtests.jl
```diff
@@ -6,6 +6,12 @@ import FiniteDifferences, ForwardDiff, Enzyme, Tracker, Zygote, ReverseDiff # ba
 import ADTypes # load support for AD types with options
 import BenchmarkTools                            # load the heuristic chunks code
 using ComponentArrays: ComponentVector           # test with other vector types
+import DifferentiationInterface
+
+DIGradient = Base.get_extension(
+    LogDensityProblemsAD,
+    :LogDensityProblemsADDifferentiationInterfaceExt
+).DIGradient
```
devmotion, line 14: Only recently I came across a Julia issue where it was argued that `Base.get_extension` is internal (something changed which broke some people's code, but I don't remember the details right now...). In any case, I think you don't need it for what you want to test - you could alternatively check `nameof(typeof(....))`.

devmotion:
```suggestion
```

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,47 @@
...
+    DIGradient <: LogDensityProblemsAD.ADGradientWrapper
+
+Gradient wrapper which uses [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl)
+
+# Fields
+
+- `backend::AbstractADType`: one of the autodiff backend types defined in [ADTypes.jl](https://github.com/SciML/ADTypes.jl), for example `ADTypes.AutoForwardDiff()`
+- `prep`: either `nothing` or the output of `DifferentiationInterface.prepare_gradient` applied to the logdensity and the provided input
+- `ℓ`: logdensity function, amenable to `LogDensityProblemsAD.logdensity(ℓ, x)`
+"""
```
devmotion, line 17: Note that the `ADGradientWrapper` types are internal and generally not documented.

test/runtests.jl
```diff
@@ -91,6 +97,9 @@ ForwardDiff.checktag(::Type{ForwardDiff.Tag{TestTag, V}}, ::Base.Fix1{typeof(log
 
     # ADTypes support
     @test typeof(ADgradient(ADTypes.AutoReverseDiff(; compile = Val(true)), ℓ)) === typeof(∇ℓ_compile)
+    @test typeof(ADgradient(ADTypes.AutoReverseDiff(; compile = Val(true)), ℓ; x=rand(3))) === typeof(∇ℓ_compile_x)
+   @test !isa(ADgradient(ADTypes.AutoReverseDiff(), ℓ), DIGradient)
+   @test !isa(ADgradient(ADTypes.AutoReverseDiff(), ℓ; x=rand(3)), DIGradient)
```
devmotion, line 102:
```suggestion
   @test nameof(typeof(ADgradient(ADTypes.AutoReverseDiff(), ℓ))) !== :DIGradient
   @test nameof(typeof(ADgradient(ADTypes.AutoReverseDiff(), ℓ; x=rand(3))) !== :DIGradient
```

test/runtests.jl
```diff
@@ -127,6 +136,8 @@ end
 
     # ADTypes support
     @test ADgradient(ADTypes.AutoForwardDiff(), ℓ) === ∇ℓ
+    @test !isa(ADgradient(ADTypes.AutoForwardDiff(), ℓ), DIGradient)
+    @test !isa(ADgradient(ADTypes.AutoForwardDiff(), ℓ; x=rand(3)), DIGradient)
```
devmotion, line 140:
```suggestion
    @test nameof(typeof(ADgradient(ADTypes.AutoForwardDiff(), ℓ))) !== :DIGradient
    @test nameof(typeof(ADgradient(ADTypes.AutoForwardDiff(), ℓ; x=rand(3)))) !== :DIGradient
```

test/runtests.jl
```diff
@@ -211,6 +223,8 @@ end
 
    # ADTypes support
    @test ADgradient(ADTypes.AutoTracker(), ℓ) === ∇ℓ
+   @test !isa(ADgradient(ADTypes.AutoTracker(), ℓ), DIGradient)
+   @test !isa(ADgradient(ADTypes.AutoTracker(), ℓ; x=rand(3)), DIGradient)
```
devmotion, line 227:
```suggestion
   @test nameof(typeof(ADgradient(ADTypes.AutoTracker(), ℓ))) !== :DIGradient
   @test nameof(typeof(ADgradient(ADTypes.AutoTracker(), ℓ; x=rand(3)))) !== :DIGradient
```

test/runtests.jl
```diff
@@ -227,6 +241,8 @@ end
 
    # ADTypes support
    @test ADgradient(ADTypes.AutoZygote(), ℓ) === ∇ℓ
+   @test !isa(ADgradient(ADTypes.AutoZygote(), ℓ), DIGradient)
+   @test !isa(ADgradient(ADTypes.AutoZygote(), ℓ; x=rand(3)), DIGradient)
```
devmotion, line 245:
```suggestion
   @test nameof(typeof(ADgradient(ADTypes.AutoZygote(), ℓ))) !== :DIGradient
   @test nameof(typeof(ADgradient(ADTypes.AutoZygote(), ℓ; x=rand(3)))) !== :DIGradient
```

test/runtests.jl
```diff
@@ -241,6 +257,8 @@ end
 
     # ADTypes support
     @test ADgradient(ADTypes.AutoEnzyme(), ℓ) === ∇ℓ_reverse
+    @test !isa(ADgradient(ADTypes.AutoEnzyme(), ℓ), DIGradient)
+    @test !isa(ADgradient(ADTypes.AutoEnzyme(), ℓ; x=rand(3)), DIGradient)
```
devmotion, line 261:
```suggestion
    @test nameof(typeof(ADgradient(ADTypes.AutoEnzyme(), ℓ))) !== :DIGradient
    @test nameof(typeof(ADgradient(ADTypes.AutoEnzyme(), ℓ; x=rand(3)))) !== :DIGradient
```

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
gdalle: In this package, the README makes it very clear that the only exposed function is `ADGradient`. Starting with Julia 1.11, we can make this information even clearer using the `public` keyword.

The remaining types and functions are internal, sure, but they can still have docstrings. It is useful for people who work on this code, or even for users who may read it while debugging. I therefore suggest we keep the docstring.

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
gdalle: A mere oversight on my end.

test/runtests.jl
gdalle: But I don't mind the change either way

ext/LogDensityProblemsADADTypesExt.jl
gdalle: I added this here and to the Zygote and Tracker fallbacks as well

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
devmotion: Yeah I'm fine with the docstring, I just wanted to mention it in case you were not aware of it.

gdalle:
Sorry I merged the review suggestions without checking typos, fixed now. The tests pass locally.

devmotion (approved):

test/runtests.jl
devmotion: It was https://github.com/JuliaLang/julia/issues/56204. As noted there, the function is neither exported nor public - but documented. It seems triage explicitly decided against exporting it: https://github.com/JuliaLang/julia/issues/47982#issuecomment-1372680535

tpapp (approved):
@gdalle, thanks for the great work, following up on all suggestions, and your patience and persistence! LGTM.

devmotion:
Yes, I think a minor release is appropriate here.

PR #46, opened 2025-06-16
version bumps, separate test project file, test fix

tpapp:
Make Enzyme tests work without using internal API.

Bump versions, extend compat.

Add a separate project file for tests.


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+# see https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"
```
devmotion, line 8: I usually run it only monthly, I tend to think that's sufficiently often for updating GH actions.

test/Project.toml
devmotion, line 1: Any particular reason to add a test/Project.toml file? In my experience, it's more stable and reliable to put test deps in the extras section in Project.toml. For instance, it can be annoying to keep packages and version specifiers in sync between Project.toml and test/Project.toml.

tpapp:
@devmotion: if you prefer I can go back to not having a `test/Project.toml`, just let me know. I want to merge this because of the compat bumps.

devmotion (approved):
I prefer having test dependencies in `[extras]` as it's easier to keep the versions of the test dependencies synced with the versions of the package dependencies (I prefer having compat entries for test dependencies as well to avoid sudden annoying test failures, but then CompatHelper will complain if non-test dependencies in test/Project.toml don't have compat entries). But I think it's fine to use a test/Project.toml in the package and only reconsider if it turns out to be problematic 🙂

PR #50, opened 2025-09-30
deprecate symbols api, clean up tests

tpapp:
Deprecate the `ADgradient(::Symbol, ...)` and `ADgradient(::Val, ...)` API. Backends should be defined using `ADTypes.jl`. **This is a breaking change and will require a major version bump once deprecations are removed**. However, I think it leads to a much cleaner API.

This PR is purposefully kept minimal and does not change the implementation unless necessary (see below). That is left for future work which this PR should make much easier, the plan is to use `DifferentiationInterface.jl` for what we can, and only do the workarounds when absolutely necessary.

However, the test framework was also cleaned up since the unified API simplifies testing. Various backends were not properly tested type stability and accepting generic vector types, and had bugs, these are now fixed.

The `shadow` kwarg from the Enzyme backend is removed since `ADTypes` does not support it. However, I don't think it was widely used, frankly, I am not even sure why it was there. The Enzyme / forward backend requires type conversions & assertions for stability, this should be investigated, but I think this PR does improve things.

I used this opportunity to deprecate the benchmarking code for ForwardDiff too, it was badly written and I don't think it belongs in this package. It should be removed.

# TODO

- [ ] update README.md
- [ ] update docs
- [ ] investigate why Mooncake does not infer (cf https://github.com/chalk-lab/Mooncake.jl/issues/792)
- [ ] save `onehot` (shadows) in Enzyme implementation
- [ ] check that all `copy` methods copy deeply when preparation is used, document (cf https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/858)


tpapp:
I just [asked](https://github.com/tpapp/LogDensityProblemsAD.jl/pull/4#issuecomment-3352143757) @wsmoses and he says we should keep them. I guess he knows best, this is beyond my understanding of Enzyme.jl.

Project.toml
```diff
@@ -19,8 +20,7 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [extensions]
-LogDensityProblemsADADTypesExt = "ADTypes"
-LogDensityProblemsADDifferentiationInterfaceExt = ["ADTypes", "DifferentiationInterface"]
+LogDensityProblemsADDifferentiationInterfaceExt = ["DifferentiationInterface"]
```
gdalle, line 23: DI is a very lightweight package (it depends on nothing except ADTypes and LinearAlgebra), I'm not sure it's worth putting it in an extension

ext/LogDensityProblemsADDifferentiationInterfaceExt.jl
```diff
@@ -1,46 +1,55 @@
...
+    least `LogDensityOrder{0}`, `backend` is used for gradient)
+    """
     ℓ::L
 end
 
-function logdensity_switched(x, ℓ)
-    # active argument must come first in DI
-    return LogDensityProblemsAD.logdensity(ℓ, x)
-end
+@inline _logdensity_callable(ℓ) = Base.Fix1(LogDensityProblems.logdensity, ℓ)
```
gdalle, line 35: The reason behind `logdensity_switched` was to avoid use of `Base.Fix1`, which is a performance pitfall for Enzyme. The `Constant` annotation on `\ell` could also help speed things up for Mooncake in the future, so I'd suggest leaving this as it was

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -49,13 +31,10 @@ function Base.show(io::IO, ∇ℓ::EnzymeGradientLogDensity)
 end
 
 function logdensity_and_gradient(∇ℓ::EnzymeGradientLogDensity{<:Any,<:Enzyme.ForwardMode},
-                                 x::AbstractVector)
-    (; ℓ, mode, shadow) = ∇ℓ
-    _shadow = shadow === nothing ? Enzyme.onehot(x) : shadow
```
gdalle, line 54: As said by Billy, this is actually useful to store ahead of time

ext/LogDensityProblemsADEnzymeExt.jl
```diff
@@ -49,13 +31,10 @@ function Base.show(io::IO, ∇ℓ::EnzymeGradientLogDensity)
...
-    ∂ℓ_∂x, y = Enzyme.autodiff(mode, logdensity, Enzyme.BatchDuplicated,
-                               Enzyme.Const(ℓ),
-                               Enzyme.BatchDuplicated(x, _shadow))
-    return y, collect(∂ℓ_∂x)
+                                 x::AbstractVector{T}) where T
+    (; ℓ, mode) = ∇ℓ
+    result = Enzyme.gradient(mode, Base.Fix1(logdensity, ℓ), x)
+    T(result.val)::T, collect(T, only(result.derivs))::Vector{T}
```
gdalle, line 37: Why convert everything to `Vector`?

gdalle (commented):
I understand why you kept the Enzyme extension, but is there a specific reason to keep the other backend extensions if you want to switch to DI? What are the things you wish you could do with DI but don't seem to be able to do?

